PR #2, opened 2017-05-26
Keyword Extraction

kevinhartman:
Adds keyword extraction to Fortis events.

Supports tagging events with single keywords and also with sequences of words (as defined by Java's word boundary tokenizer).

For example, "am so glad" will be found in the following:
"I am so glad my cat likes his cat food."

But not in:
"Iam so glad my cat likes his food."

See test spec for more examples of extraction behavior.

### DIFF
--- a/src/main/scala/DemoFortis.scala
+++ b/src/main/scala/DemoFortis.scala
@@ -10,6 +10,7 @@ import com.microsoft.partnercatalyst.fortis.spark.transforms.locations.client.Fe
 import com.microsoft.partnercatalyst.fortis.spark.transforms.locations.nlp.PlaceRecognizer
 import com.microsoft.partnercatalyst.fortis.spark.transforms.locations.{Geofence, LocationsExtractor}
 import com.microsoft.partnercatalyst.fortis.spark.transforms.sentiment.{SentimentDetector, SentimentDetectorAuth}
+import com.microsoft.partnercatalyst.fortis.spark.transforms.topic.KeywordExtractor
 import org.apache.log4j.{Level, Logger}
 import org.apache.spark.streaming.{Seconds, StreamingContext}
 import org.apache.spark.{SparkConf, SparkContext}
@@ -57,6 +58,7 @@ object DemoFortis {
     val placeRecognizer = new PlaceRecognizer(Option(System.getenv("FORTIS_MODELS_DIRECTORY")))
     val featureServiceClient = new FeatureServiceClient(System.getenv("FORTIS_FEATURE_SERVICE_HOST"))
     val locationsExtractor = new LocationsExtractor(featureServiceClient, geofence, Some(placeRecognizer)).buildLookup()
+    val keywordExtractor = new KeywordExtractor(List("Ariana"))
     val imageAnalysis = new ImageAnalyzer(ImageAnalysisAuth(System.getenv("OXFORD_VISION_TOKEN")), featureServiceClient)
     val languageDetection = new LanguageDetector(LanguageDetectorAuth(System.getenv("OXFORD_LANGUAGE_TOKEN")))
     val sentimentDetection = new SentimentDetector(SentimentDetectorAuth(System.getenv("OXFORD_LANGUAGE_TOKEN")))
@@ -68,7 +70,8 @@ object DemoFortis {
           .map(instagram => {
             // do computer vision analysis: keyword extraction, etc.
             val source = instagram.link
-            val analysis = imageAnalysis.analyze(instagram.images.standard_resolution.url)
+            var analysis = imageAnalysis.analyze(instagram.images.standard_resolution.url)
+            analysis = analysis.copy(keywords = keywordExtractor.extractKeywords(instagram.caption.text))
             AnalyzedItem(originalItem = instagram, analysis = analysis, source = source)
           })
           .map(analyzedInstagram => {
@@ -93,7 +96,7 @@ object DemoFortis {
           .map(tweet => {
             val source = s"https://twitter.com/statuses/${tweet.getId}"
             val language = if (Option(tweet.getLang).isDefined) { Option(tweet.getLang) } else { languageDetection.detectLanguage(tweet.getText) }
-            val analysis = Analysis(language = language)
+            val analysis = Analysis(language = language, keywords = keywordExtractor.extractKeywords(tweet.getText))
             AnalyzedItem(originalItem = tweet, analysis = analysis, source = source)
           })
           .filter(analyzedPost => {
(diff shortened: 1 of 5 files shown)


src/main/scala/DemoFortis.scala
```diff
@@ -57,6 +58,7 @@ object DemoFortis {
     val placeRecognizer = new PlaceRecognizer(Option(System.getenv("FORTIS_MODELS_DIRECTORY")))
     val featureServiceClient = new FeatureServiceClient(System.getenv("FORTIS_FEATURE_SERVICE_HOST"))
     val locationsExtractor = new LocationsExtractor(featureServiceClient, geofence, Some(placeRecognizer)).buildLookup()
+    val keywordExtractor = new KeywordExtractor(List("Ariana"))
```
c-w, line 61: Note for the future: we may want to broadcast all of these helpers so that we avoid sending a copy of them over the network for every task and instead just send one copy for every worker and have it cached. [More info.](https://jaceklaskowski.gitbooks.io/mastering-apache-spark/content/spark-broadcast.html)

src/main/scala/DemoFortis.scala
```diff
@@ -134,7 +137,7 @@ object DemoFortis {
           .map(post => {
             val source = post.post.getPermalinkUrl.toString
             val language = languageDetection.detectLanguage(post.post.getMessage)
-            val analysis = Analysis(language = language)
+            val analysis = Analysis(language = language, keywords = keywordExtractor.extractKeywords(post.post.getMessage))
```
c-w, line 140: We may want to do the keyword extractions as a separate map step instead of merged with another transform. This has a few advantages:
1. Conceptually simpler: every map just does one thing
2. We can more easily switch different processing steps on or off by just removing a map call
3. The map operations are independent so can be done in parallel

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/topic/KeywordExtractor.scala
```diff
@@ -0,0 +1,38 @@
...
+import com.microsoft.partnercatalyst.fortis.spark.transforms.Tag
+import org.apache.commons.collections4.trie.PatriciaTrie
+
+import scala.collection.mutable.ListBuffer
+
+@SerialVersionUID(100L)
+class KeywordExtractor(keyWords: Seq[String]) extends Serializable {
+  private val wordTokenizer = """\b""".r
+  private val keywordTrie = new PatriciaTrie[Unit]()
+  private val originalCase = Map[String, String](keyWords.map(s => (s.toLowerCase, s)): _*)
```
c-w: These could be lazy and transient so that we rebuild them on the worker instead of sending them over the network (likely faster).

c-w: This map could potentially be large. Do we really care about preserving the case?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/topic/KeywordExtractor.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  originalCase.keys.foreach(keywordTrie.put(_, ()))
+
+  def extractKeywords(text: String): List[Tag] = {
+
+    def findMatches(segment: Seq[String]): Iterable[String] = {
```
c-w: Note: unlike Javascript or Python, having nested functions is _not_ a performance hit in Scala. During the compile step, the inner function gets lifted to the class level so we don't re-declare it on every method call like we'd do in the dynamic languages. [Read more.](https://stackoverflow.com/a/2482410/3817588) So it's all good to use the inner functions :)

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/topic/KeywordExtractor.scala
```diff
@@ -0,0 +1,38 @@
...
+        if (keywordTrie.containsKey(prefix)) {
+          result.append(originalCase(prefix))
+        }
+      }
+
+      result
+    }
+
+    val tokens = wordTokenizer.split(text.toLowerCase).toSeq
+    tokens.tails.flatMap(findMatches(_).map(Tag(_, 1))).toList
```
c-w, line 31: Why do we always want to ignore the first token?

c-w (approved):
LGTM.

src/main/scala/DemoFortis.scala
kevinhartman: Makes sense. The other thing is we should be filtering the results based on the presence of keywords. I just put this here as a quick POC.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/topic/KeywordExtractor.scala
kevinhartman: Great idea.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/topic/KeywordExtractor.scala
kevinhartman: To clarify, is this just a note for other readers, or is there some recommendation here?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/topic/KeywordExtractor.scala
kevinhartman: The first item in "tails" will be the entire collection :) See https://www.scala-lang.org/api/current/scala/collection/Seq.html#tails:Iterator[Repr]

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/topic/KeywordExtractor.scala
kevinhartman: Heh, actually I should just be storing the originally cased version as the Trie values if at all. @erikschlegel, do we want to preserve the casing of keywords when writing AnalyzedItems?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/topic/KeywordExtractor.scala
c-w: That's a note for others. It's something I learned while reviewing this code so I figured it'd be good to share it.

PR #3, opened 2017-06-03
Support for Tadaweb over EventHub

kevinhartman:


src/main/scala/DemoFortis.scala
```diff
@@ -173,6 +181,66 @@ object DemoFortis {
...
+            analysis = Analysis(keywords =
+              keywordExtractor.extractKeywords(event.text) ::: keywordExtractor.extractKeywords(event.title)
+            )
+          ))
+          .filter(_.analysis.keywords.nonEmpty)
+          .map(fortisEvent => {
+            val language = languageDetection.detectLanguage(fortisEvent.originalItem.text)
+            val sentiment: Option[Double] = fortisEvent.originalItem.sentiment match {
+              case "negative" => Some(0)
+              case "neutral" => Some(0.6)
```
c-w, line 199: Wouldn't "neutral" be 0.5 in the cognitive service system?

src/main/scala/DemoFortis.scala
```diff
@@ -173,6 +181,66 @@ object DemoFortis {
...
+              case "positive" => Some(1)
+              case _ => language match {
+                case Some(lang) =>
+                  if (supportedLanguages.contains(lang))
```
c-w: If this is here to ensure that the sentiment detector can handle the language of the event, it already does this check for you, so no need to do it here ([source](https://github.com/CatalystCode/project-fortis-spark/blob/23b90f13750fff59bebb6f9d87346579b46b16df/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/SentimentDetector.scala#L16)). If we want to filter out events that aren't from our target languages, I'd say to do it as a separate step to increase clarity (so one `.map` to add the language followed by a `.filter` to remove the non-targeted events, followed by a `.map` to add the sentiment).

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/streamfactories/EventHubStreamFactory.scala
```diff
@@ -5,15 +5,20 @@ import org.apache.spark.streaming.StreamingContext
...
 
+import java.nio.charset.StandardCharsets
+import scala.language.implicitConversions
 import scala.reflect.ClassTag
+import scala.util.{Failure, Success, Try}
 
-class EventHubStreamFactory[A: ClassTag](
-    progressDir: String,
-    adapter: Array[Byte] => A)
+class EventHubStreamFactory[A: ClassTag](identifier: String, adapter: (Array[Byte]) => Try[A], progressDir: String)
```
c-w, line 13: Is the `progressDir` meaningful for clients of this code or can it just be a temp directory to simplify the constructor?

src/main/scala/DemoFortis.scala
```diff
@@ -173,6 +181,66 @@ object DemoFortis {
       }
     }
 
+    if (mode.contains("tadaweb")) {
```
c-w, line 184: These demos are getting quite many. We may want to split them out into separate classes.

c-w (approved):
LGTM.

src/main/scala/DemoFortis.scala
kevinhartman: That's my impression as well, but as per a conversation with @erikschlegel, the correct value here is 0.6. Erik, please let us know if this is correct in context.

src/main/scala/DemoFortis.scala
kevinhartman: Removed check. Since the sentiment detector is able to handle the unsupported case, it seems cleaner to me to filter as the next stage after attempting to detect sentiment. For Tadaweb, the sentiment can come from the event itself. Filtering in this way avoids duplicating that check.

kevinhartman: That said, I will also add a filter to prevent items without a sentiment from flowing onward.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/streamfactories/EventHubStreamFactory.scala
kevinhartman: I think it's important that it's configurable since it's expected to be an HDFS share in production.

src/main/scala/DemoFortis.scala
kevinhartman: I would agree. Let's save that for a separate PR. Depending on how checkpointing configuration looks, the task for adding that might be an ideal one for moving this code around.

src/main/scala/DemoFortis.scala
erikschlegel: I originally wanted to slightly lean towards the positive side for neutrality, but the more I think about it we should be consistent with cognitive services.

erikschlegel (commented):
LGTM. Nice work @kevinhartman

PR #8, opened 2017-06-06
Implement radio ingestion

c-w:
This PR implements [Audio Streaming Integration](https://github.com/CatalystCode/project-fortis/issues/45) by integrating the [SpeechToText-WebSockets-Java](https://github.com/CatalystCode/SpeechToText-WebSockets-Java) with our Spark platform.

The integration works as follows:
- We create a transcriber object via the speech-to-text library.
- We open a standard URL connection to the radio stream and connect it to the transcriber.
- The transcriber takes care of reading the audio stream, converting it to text, etc.
- Whenever a transcription is complete, the transcriber calls our callback. In this callback, we store the transcribed audio in our Spark receiver which triggers the downstream Spark Streaming pipeline.

Pipeline in action:

image

Note that the speech-to-text library is still fairly brittle and will need some improvements to be fully production-ready. I recommend to start playing with the integration, collecting some issues and then fixing them in bulk.

The feature service seems to be broken/changed recently, so to test the core of this PR, I also needed to make our feature service client more robust to failures. These changes are also included in this PR: we now ignore bad responses from the feature service and we also ignore when we're unable to connect to the feature service. Even if the feature service is down, the rest of the pipeline can still work (just the locations-from-text inference piece won't work) so I assert that it's good to make our pipeline robust to feature service outages.


src/main/scala/DemoFortis.scala
```diff
@@ -98,6 +103,36 @@ object DemoFortis {
...
+            AnalyzedItem(originalItem = radioTranscription, analysis = analysis, source = source)
+          })
+          .filter(analyzedRadio => {
+            supportedLanguages.contains(analyzedRadio.analysis.language.getOrElse(""))
+          })
+          .map(analyzedRadio => {
+            // sentiment detection
+            val text = analyzedRadio.originalItem.text
+            val language = analyzedRadio.analysis.language.getOrElse("")
+            val inferredSentiment = sentimentDetection.detectSentiment(text, language).map(List(_)).getOrElse(List())
```
kevinhartman, line 124: `toList` on `Option[String]` should do what you want :)

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/streamfactories/RadioStreamFactory.scala
```diff
@@ -0,0 +1,96 @@
...
+  speechType: String,
+  outputFormat: String,
+  storageLevel: StorageLevel
+) extends Receiver[RadioTranscription](storageLevel) {
+
+  private val language = new Locale(locale).getLanguage
+  private var audioStream: InputStream = _
+  private var transcriber: Transcriber = _
+
+  private lazy val onTranscription = new Func[String] {
```
kevinhartman: Perhaps a comment for another repo, but why is there a custom class for Func coming from speechtotext.utils?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/streamfactories/RadioStreamFactory.scala
```diff
@@ -0,0 +1,96 @@
...
+    transcriber.transcribe(audioStream, onTranscription, onHypothesis)
+  }
+
+  override def onStop(): Unit = {
+    if (audioStream != null) audioStream.close()
+    if (transcriber != null) transcriber = null
+  }
+}
+
+case class RadioTranscription(
```
kevinhartman: I'd move this out to a dto folder. I can imagine clients that may want to work with RadioTranscriptions but not RadioStreamFactory.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/streamfactories/RadioStreamFactory.scala
```diff
@@ -0,0 +1,96 @@
...
+      val speechType = params("speechType")
+      val outputFormat = params("outputFormat")
+
+      new RadioInputDStream(
+        ssc, radioUrl, audioType, locale, subscriptionKey,
+        speechType, outputFormat, StorageLevel.MEMORY_ONLY)
+  }
+}
+
+class RadioInputDStream(
```
kevinhartman: I think we should move this class and its dependencies into a separate package for cleaner separation.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/client/FeatureServiceClient.scala
```diff
@@ -23,21 +26,37 @@ class FeatureServiceClient(host: String) extends Serializable {
   private def parseResponse(response: String): Iterable[FeatureServiceFeature] = {
     implicit val formats = json.DefaultFormats
 
-    json.parse(response).extract[FeatureServiceResponse].features
+    try {
+      json.parse(response).extract[FeatureServiceResponse].features
+    } catch {
+      case ex: ParseException =>
+        logError(s"Unable to parse feature service response: $response", ex)
+        List()
```
kevinhartman: As a client of this code, I'd like to distinguish between the error case and having no features found. Perhaps this interface would be a good candidate for a return type of Try[T]?

[Here's a pretty good article](http://danielwestheide.com/blog/2012/12/26/the-neophytes-guide-to-scala-part-6-error-handling-with-try.html) describing functional error handling with Try in Scala.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/client/FeatureServiceClient.scala
```diff
@@ -23,21 +26,37 @@ class FeatureServiceClient(host: String) extends Serializable {
...
+    fetchResponse(fetch)
   }
 
   protected def fetchNameResponse(names: Iterable[String]): String = {
     val fetch = s"http://$host/features/name/${names.mkString(",")}"
-    Source.fromURL(fetch)("UTF-8").mkString
+    fetchResponse(fetch)
+  }
+
+  private def fetchResponse(url: String): String = {
```
kevinhartman: This looks like another good candidate for Try[String].

kevinhartman (approved):
LGTM

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/streamfactories/RadioStreamFactory.scala
c-w: `speechtotext.utils.Func` is part of the Java speech-to-text websocket package. Java doesn't have a native function interface so I had to make one.

src/main/scala/DemoFortis.scala
c-w: Ugh, Scala strikes again -- that's not what I'd call more readable/understandable. `getOrElse` seems more explicit to me.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/streamfactories/RadioStreamFactory.scala
kevinhartman: I believe this is intended to fill that gap:
https://docs.oracle.com/javase/8/docs/api/java/util/function/Consumer.html

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/client/FeatureServiceClient.scala
c-w: Done in c110efa.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/client/FeatureServiceClient.scala
c-w: Done in c110efa.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/streamfactories/RadioStreamFactory.scala
c-w: Done in 33347bb

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/streamfactories/RadioStreamFactory.scala
c-w: Done in 33347bb

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/streamfactories/RadioStreamFactory.scala
c-w: It shows that most of my Java-days were pre-8. I'll update the STT library and push a new version.

src/main/scala/DemoFortis.scala
kevinhartman: Heh yeah but I think the major offense is just that you can enumerate an Option. Since we're already doing that, toList doesn't really push the boundaries too much further IMO :)

c-w: Currently we use `getOrElse` in many places and `isDefined / get` in others. We should in bulk move all of this to whichever approach you think is better going forward (e.g. `match / case`).

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/client/FeatureServiceClient.scala
```diff
@@ -1,43 +1,59 @@
...
+    unpack(fetchPointResponse(latitude = latitude, longitude = longitude), "point")
   }
 
   def name(names: Iterable[String]): Iterable[FeatureServiceFeature] = {
-    parseResponse(fetchNameResponse(names))
+    unpack(fetchNameResponse(names), "name")
   }
 
-  private def parseResponse(response: String): Iterable[FeatureServiceFeature] = {
+  private def unpack(responseBody: Try[String], endpointName: String): Iterable[FeatureServiceFeature] = {
```
kevinhartman, line 24: The error is still swallowed here. I would expect that clients of the public methods of this class should be able to distinguish between empty results and an internal error.

src/main/scala/DemoFortis.scala
kevinhartman: SGTM

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/client/FeatureServiceClient.scala
c-w: Right now is not the time to change the public API of this class. We currently have many cases where we're not handling dependency failure so we need to audit this anyways in the future.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/streamfactories/RadioStreamFactory.scala
c-w: Done in 23d6409.

c-w:
Addressed all actionable comments from Kevin, now waiting for sign off from @erikschlegel.

src/main/scala/DemoFortis.scala
```diff
@@ -278,6 +314,19 @@ object DemoFortis {
           )
         )
       ),
+      "radio" -> List(
+        ConnectorConfig(
+          "Radio",
```
erikschlegel, line 319: Can you provide an example with streaming from more than one source, just to prove out that we can support a list of audio sources.

erikschlegel (approved):
Code looks great @c-w . Only recommendation was to include one other radio stream to prove out that use-case. 

Great work!

src/main/scala/DemoFortis.scala
c-w: The radio stream is designed to process a single radio feed. For ingesting multiple radio feeds, we should use multiple radio streams. I'm syncing up with @kevinhartman to see how we'd represent that in the current stream-factory paradigm since your ask is really a question for the stream-factories rather than the stream implementations.

c-w:
Closing this off. Let's revisit the questions around multiple copies of the same stream configured with different parameters as a separate thread.

PR #9, opened 2017-06-07
Add sentiment analysis capability for 68 more languages

c-w:
In the past, we integrated with the Cognitive Services Text Analytics API to extract the sentiment from a text. When we started the integration, the API only had support for 4 languages. Now, there is support for 15 languages.† However, Fortis would like to support many more languages. This PR addresses that problem.

Most sentiment detection work is done for the big NLP languages (English, German, etc.), however, I did manage to find a [paper](http://www.aclweb.org/anthology/P14-2063) by the data science group at Stony Brook University which focused on sentiment analysis work for all the other languages out there. Through a combination of a variety of techniques (including machine translation and morphological propagation across linguistically similar languages), they managed to create word-polarity [lists](https://sites.google.com/site/datascienceslab/projects/multilingualsentiment) for over 100 languages. I analyzed the lists they provide, kept the lists for the languages where they have at least 500 positive and negative terms and uploaded them to our fortis-models blob in a machine readable format.

Inside of Fortis, we then use the word polarity lists to compute sentiment for languages that are unsupported by Cognitive Services, like so:

1) Download the polarity lists from blob (cache on disk)
2) Read the positive/negative words (cache in memory)
3) Count how many positive/negative words are in the sentence.
4) If there are more positive words, assume positive sentiment. If there are more negative words, assume negative sentiment. If the count is the same, assume neutral sentiment.

This approach is super naive (I've asked for support from the Machine Learning TWG to improve the approach if possible), but it's better than nothing and similar to how we did sentiment analysis in Fortis-v1 so I assert it's good enough for now.

This closes [Issue#18](https://github.com/CatalystCode/project-fortis/issues/18).

NB: The PR also includes some re-structuring that was necessary to implement the new functionality, e.g. moving around some packages, extracting helper utilities, etc.

†: Cognitive Services currently support English, Spanish, Portuguese, French, German, Italian, Dutch, Norwegian, Swedish, Polish, Danish, Finnish, Russian, Greek and Turkish.

### DIFF
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/ZipModelsProvider.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/ZipModelsProvider.scala
@@ -13,7 +13,7 @@ import scala.sys.process._
 
 @SerialVersionUID(100L)
 class ZipModelsProvider(
-  formatModelsDownloadUrl: String => String,
+  modelsUrlFromLanguage: String => String,
   modelsSource: Option[String] = None
 ) extends Serializable with Loggable {
 
@@ -33,7 +33,7 @@ class ZipModelsProvider(
       return previouslyDownloadedPath
     }
 
-    val remotePath = modelsSource.getOrElse(formatModelsDownloadUrl(language))
+    val remotePath = modelsSource.getOrElse(modelsUrlFromLanguage(language))
     if ((!remotePath.startsWith("http://") && !remotePath.startsWith("https://")) || !remotePath.endsWith(".zip")) {
       throw new FileNotFoundException(s"Unable to process $remotePath, should be http(s) link to zip file")
     }
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/ZipModelsProvider.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/ZipModelsProvider.scala
@@ -5,7 +5,7 @@ import java.net.URL
 import java.nio.file.Files
 import java.util.concurrent.ConcurrentHashMap
 
-import com.microsoft.partnercatalyst.fortis.spark.logging.Logger
+import com.microsoft.partnercatalyst.fortis.spark.logging.Loggable
 import net.lingala.zip4j.core.ZipFile
 
 import scala.collection.JavaConversions._
@@ -15,7 +15,7 @@ import scala.sys.process._
 class ZipModelsProvider(
   formatModelsDownloadUrl: String => String,
   modelsSource: Option[String] = None
-) extends Serializable with Logger {
+) extends Serializable with Loggable {
 
   @volatile private lazy val modelDirectories = new ConcurrentHashMap[String, String]
 
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/ZipModelsProvider.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/ZipModelsProvider.scala
@@ -5,7 +5,7 @@ import java.net.URL
 import java.nio.file.Files
 import java.util.concurrent.ConcurrentHashMap
 
-import com.microsoft.partnercatalyst.fortis.spark.transforms.locations.Logger
+import com.microsoft.partnercatalyst.fortis.spark.logging.Logger
 import net.lingala.zip4j.core.ZipFile
 
 import scala.collection.JavaConversions._
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/ZipModelsProvider.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/ZipModelsProvider.scala
@@ -12,15 +12,14 @@ import scala.collection.JavaConversions._
 import scala.sys.process._
 
 @SerialVersionUID(100L)
-abstract class HasZipModels(
+class ZipModelsProvider(
+  formatModelsDownloadUrl: String => String,
   modelsSource: Option[String] = None
 ) extends Serializable with Logger {
 
   @volatile private lazy val modelDirectories = new ConcurrentHashMap[String, String]
 
-  protected def formatModelsDownloadUrl(language: String): String
-
-  protected def ensureModelsAreDownloaded(language: String): String = {
+  def ensureModelsAreDownloaded(language: String): String = {
     val localPath = modelsSource.getOrElse("")
     if (hasModelFiles(localPath, language)) {
       logDebug(s"Using locally provided model files from $localPath")
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/PlaceRecognizer.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/PlaceRecognizer.scala
@@ -2,7 +2,7 @@ package com.microsoft.partnercatalyst.fortis.spark.transforms.locations
 
 import java.io.IOError
 
-import com.microsoft.partnercatalyst.fortis.spark.logging.Logger
+import com.microsoft.partnercatalyst.fortis.spark.logging.Loggable
 import com.microsoft.partnercatalyst.fortis.spark.transforms.ZipModelsProvider
 import com.microsoft.partnercatalyst.fortis.spark.transforms.nlp.OpeNER
 import ixa.kaflib.Entity
@@ -13,7 +13,7 @@ import scala.collection.JavaConversions._
 class PlaceRecognizer(
   modelsSource: Option[String] = None,
   enabledLanguages: Set[String] = Set("de", "en", "es", "eu", "it", "nl")
-) extends Serializable with Logger {
+) extends Serializable with Loggable {
 
   @volatile private lazy val modelsProvider = createModelsProvider()
 
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/PlaceRecognizer.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/PlaceRecognizer.scala
@@ -1,6 +1,6 @@
 package com.microsoft.partnercatalyst.fortis.spark.transforms.locations
 
-import java.io.IOError
+import java.io.{IOError, IOException}
 
 import com.microsoft.partnercatalyst.fortis.spark.logging.Loggable
 import com.microsoft.partnercatalyst.fortis.spark.transforms.ZipModelsProvider
@@ -8,6 +8,7 @@ import com.microsoft.partnercatalyst.fortis.spark.transforms.nlp.OpeNER
 import ixa.kaflib.Entity
 
 import scala.collection.JavaConversions._
+import scala.util.{Failure, Success, Try}
 
 @SerialVersionUID(100L)
 class PlaceRecognizer(
@@ -22,9 +23,18 @@ class PlaceRecognizer(
       return Set()
     }
 
-    try {
-      val resourcesDirectory = modelsProvider.ensureModelsAreDownloaded(language)
+    Try(modelsProvider.ensureModelsAreDownloaded(language)) match {
+      case Failure(ex) =>
+        logError(s"Unable to load models for language $language", ex)
+        Set()
 
+      case Success(resourcesDirectory) =>
+        extractPlacesUsingModels(text, language, resourcesDirectory)
+    }
+  }
+
+  private def extractPlacesUsingModels(text: String, language: String, resourcesDirectory: String): Set[String] = {
+    try {
       val kaf = OpeNER.tokAnnotate(resourcesDirectory, text, language)
       OpeNER.posAnnotate(resourcesDirectory, language, kaf)
       OpeNER.nerAnnotate(resourcesDirectory, language, kaf)
@@ -33,7 +43,7 @@ class PlaceRecognizer(
 
       kaf.getEntities.toList.filter(entityIsPlace).map(_.getStr).toSet
     } catch {
-      case ex @ (_ : NullPointerException | _ : IOError) =>
+      case ex @ (_ : NullPointerException | _ : IOError | _ : IOException) =>
         logError(s"Unable to extract places for language $language", ex)
         Set()
     }
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/PlaceRecognizer.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/PlaceRecognizer.scala
@@ -14,7 +14,7 @@ class PlaceRecognizer(
   enabledLanguages: Set[String] = Set("de", "en", "es", "eu", "it", "nl")
 ) extends Serializable with Logger {
 
-  @volatile private lazy val modelsProvider = new ZipModelsProvider(formatModelsDownloadUrl, modelsSource)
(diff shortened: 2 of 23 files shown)


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/SentimentDetector.scala
```diff
@@ -1,49 +1,31 @@
...
+        Some(sentimentScore)
+      case Success(None) | Failure(_) =>
+        logDebug(s"Unable to compute sentiment via cognitive services, falling back to word-list approach for $language")
+        wordlistSentimentDetector.detectSentiment(text, language)
     }
   }
 }
+
+object SentimentDetector {
+  val POSITIVE: Double = 1.0
```
kevinhartman: Scala constants should be Pascal-cased.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/ZipModelsProvider.scala
```diff
@@ -1,53 +1,25 @@
...
 import scala.collection.JavaConversions._
 import scala.sys.process._
 
 @SerialVersionUID(100L)
-class PlaceRecognizer(
-  modelsSource: Option[String] = None,
-  enabledLanguages: Set[String] = Set("de", "en", "es", "eu", "it", "nl")
-) extends Serializable with Logger {
+class ZipModelsProvider(
+  formatModelsDownloadUrl: String => String,
```
kevinhartman: Naming this like a function might make its use clearer. Maybe something like `modelsUrlFromLanguage`.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/ZipModelsProvider.scala
```diff
@@ -1,53 +1,25 @@
...
+  modelsSource: Option[String] = None
```
kevinhartman: nit: how about `modelDirectory`?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/PlaceRecognizer.scala
```diff
@@ -0,0 +1,52 @@
...
+    try {
+      val resourcesDirectory = modelsProvider.ensureModelsAreDownloaded(language)
+
+      val kaf = OpeNER.tokAnnotate(resourcesDirectory, text, language)
+      OpeNER.posAnnotate(resourcesDirectory, language, kaf)
+      OpeNER.nerAnnotate(resourcesDirectory, language, kaf)
+
+      logDebug(s"Analyzed text $text in language $language: $kaf")
+
+      kaf.getEntities.toList.filter(entityIsPlace).map(_.getStr).toSet
```
kevinhartman, line 44: `toSet` not needed here since return type is `Iterable`.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/PlaceRecognizer.scala
```diff
@@ -0,0 +1,52 @@
...
+    } catch {
+      case ex @ (_ : NullPointerException | _ : IOError) =>
```
kevinhartman: Which code will throw a null pointer exception? If it's OpeNER, we should likely be calling it differently, since throwing NullPointerException should not be part of an error handling contract.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/SentimentDetector.scala
```diff
@@ -1,49 +1,31 @@
...
 
-  protected def parseResponse(apiResponse: String, textId: String): Option[Double] = {
-    implicit val formats = json.DefaultFormats
-    val response = json.parse(apiResponse).extract[dto.JsonSentimentDetectionResponse]
-    if (response.errors.exists(_.id == textId)) {
-      None
-    } else {
-      response.documents.find(_.id == textId).map(_.score)
+  def detectSentiment(text: String, language: String): Option[Double] = {
+    val sentiment = Try(cognitiveServicesSentimentDetector.detectSentiment(text, language))
```
kevinhartman: I'd rather see `detectSentiment` return a `Try` so that its contract includes the possibility of external failure. Otherwise, from the caller's perspective, it's not clear that error handling is required.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/WordListSentimentDetector.scala
```diff
@@ -0,0 +1,77 @@
...
+
+import com.microsoft.partnercatalyst.fortis.spark.logging.Loggable
+import com.microsoft.partnercatalyst.fortis.spark.transforms.ZipModelsProvider
+import com.microsoft.partnercatalyst.fortis.spark.transforms.nlp.Tokenizer.tokenize
+import com.microsoft.partnercatalyst.fortis.spark.transforms.sentiment.SentimentDetector.{NEGATIVE, NEUTRAL, POSITIVE}
+
+import scala.io.Source
+
+@SerialVersionUID(100L)
+class WordListSentimentDetector(
```
kevinhartman, line 13: Perhaps you could define a trait for SentimentDetector.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/WordListSentimentDetector.scala
```diff
@@ -0,0 +1,77 @@
+package com.microsoft.partnercatalyst.fortis.spark.transforms.sentiment
+
+import java.io.{File, IOError, IOException}
+import java.util.concurrent.ConcurrentHashMap
+
+import com.microsoft.partnercatalyst.fortis.spark.logging.Loggable
+import com.microsoft.partnercatalyst.fortis.spark.transforms.ZipModelsProvider
+import com.microsoft.partnercatalyst.fortis.spark.transforms.nlp.Tokenizer.tokenize
```
kevinhartman: Importing a specific function like this globally seems a little misleading. If you change `tokenize` to `apply` in `Tokenizer`, then you can tokenize in this file with `Tokenizer("input string")`

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/nlp/Tokenizer.scala
```diff
@@ -0,0 +1,9 @@
+package com.microsoft.partnercatalyst.fortis.spark.transforms.nlp
+
+object Tokenizer {
+  @transient private lazy val wordTokenizer = """\b""".r
+
+  def tokenize(sentence: String): Seq[String] = {
```
kevinhartman: Might as well call this `apply` so we can treat Tokenizer like a global function.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/WordListSentimentDetector.scala
```diff
@@ -0,0 +1,77 @@
...
+      computeSentimentScore(numPositiveWords, numNegativeWords)
+    } catch {
+      case ex @ (_ : IOException | _ : IOError) =>
+        logError(s"Unable to extract sentiment for language $language", ex)
+        None
+    }
+  }
+
+  private def computeSentimentScore(numPositiveWords: Int, numNegativeWords: Int) = {
+    if (numPositiveWords > numNegativeWords) {
```
kevinhartman, line 35: This would be much more beautiful with a `match` :)

src/test/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/WordListSentimentDetectorSpec.scala
```diff
@@ -0,0 +1,66 @@
...
+}
+
+class TestWordListSentimentDetector(
+  positiveWords: Set[String],
+  negativeWords: Set[String],
+  modelsProvider: ZipModelsProvider = new TestZipModelsProvider
+) extends WordListSentimentDetector {
+
+  protected override def readWords(path: String): Set[String] = {
+    if (path.contains("pos.txt")) {
```
kevinhartman: nit: endsWith

kevinhartman (approved):
LGTM w/ comments.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/SentimentDetector.scala
c-w: Sure; however, we currently don't do error handling anywhere, so this is best left as a productionization task in bulk.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/PlaceRecognizer.scala
c-w: I wish it weren't the case... but unfortunately the OpeNER code is research-quality code so it throws a NPE in some cases when it can't load a model since they swallow an IOException internally and then later hit the null-pointer.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/WordListSentimentDetector.scala
c-w: I'm not a fan of sprinkling Scala constructs left right and center which make the code harder to understand for folks coming from more mainstream languages.

src/test/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/WordListSentimentDetectorSpec.scala
c-w: Done in d5cbd60.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/nlp/Tokenizer.scala
c-w: Done in 880c5af

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/WordListSentimentDetector.scala
c-w: Done in 880c5af

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/PlaceRecognizer.scala
c-w: The function fetches the entities in a sentence so there shouldn't be duplicates. I've changed the return type to Set in 5861eab.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/ZipModelsProvider.scala
c-w: Done in 9ec3011

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/SentimentDetector.scala
c-w: Done in cb10a83

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/ZipModelsProvider.scala
c-w: It could also be a custom URL to a models zip file.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/WordListSentimentDetector.scala
c-w: There could be an interface for it, but right now there's only two implementations which are both wrapped by a single class so there wouldn't be much value in it.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/SentimentDetector.scala
kevinhartman: I think we should do this now since this is new code. We're doing error handling in other components (stream provider, for example).

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/PlaceRecognizer.scala
kevinhartman: Ah yikes. Can we tighten the Try scope to just the relevant OpeNER statements?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/WordListSentimentDetector.scala
kevinhartman: I agree that it wouldn't provide immediate value, but might be nice to think through to set us up for easier extensibility in the future.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/WordListSentimentDetector.scala
kevinhartman: Looking at this again, I'd say it's fine as is, but as a general response, this codebase is written in Scala 😛.

c-w: Sure; then again, Java is as close to a lingua franca as we get in the software world so using Scala as Java++ (as opposed to going crazy and using Scala as a Haskell dialect) will make the code easier for others to grok :)

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/WordListSentimentDetector.scala
c-w: Done in c7008f8.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/PlaceRecognizer.scala
c-w: Done in e89573a.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/sentiment/SentimentDetector.scala
c-w: Let's do a review of the transforms for error-handling as a follow up. That'll be a more streamlined use of time, will avoid one of my favorite anti-patterns ([lava-layer](http://mikehadlow.blogspot.de/2014/12/the-lava-layer-anti-pattern.html)) and will fit nicely with moving away from the demo Spark job as we'll also be able to generalize our stream processing at that point (e.g. right now there's lots of copy-pasting in the pipeline processing code which we can rationalize and add error handling).

PR #11, opened 2017-06-15
Enable checkpointing and refactor main

kevinhartman:
- Enables checkpointing
- Refactors `DemoFortis` into `ProjectFortis` and companion files.

See TODO comments in `ProjectFortis.scala` for future iteration details.


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/dto/FortisItem.scala
```diff
@@ -0,0 +1,7 @@
+package com.microsoft.partnercatalyst.fortis.spark.dto
+
+trait FortisItem {
```
c-w, line 3: Why do we need interfaces for which there will only ever be one implementation? We can always introduce the interface at the point at which we'll need it. Less code is always better as it's less overhead to understand and maintain; [yagni](https://martinfowler.com/bliki/Yagni.html). Same applies to `TransformContext` below.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/StreamProviderFactory.scala
```diff
@@ -0,0 +1,43 @@
+package com.microsoft.partnercatalyst.fortis.spark
+
+import java.io.File
+
+import com.microsoft.partnercatalyst.fortis.spark.streamfactories.adapters.TadawebAdapter
+import com.microsoft.partnercatalyst.fortis.spark.streamfactories._
+import com.microsoft.partnercatalyst.fortis.spark.streamprovider.StreamProvider
+
+object StreamProviderFactory {
+  def create()(implicit settings: Settings): StreamProvider = {
```
c-w, line 10: What's wrong with just passing in the settings via an argument? It's easier to understand (so easier to onboard people) and we don't run the risk of [over-using implicits](https://stackoverflow.com/a/41709221/3817588) which happens very quickly... Generally speaking, the fewer fancy Scala feature we can use when they're avoidable, the happier I'd be: [simpler code is better code](https://chrismm.com/blog/writing-good-code-reduce-the-cognitive-load/).

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
```diff
@@ -0,0 +1,194 @@
...
+
+    val progressDir: String = envOrFail(Constants.Env.HighlyAvailableProgressDir)
+    val featureServiceHost: String = envOrFail(Constants.Env.FeatureServiceHost)
+    val oxfordLanguageToken: String = envOrFail(Constants.Env.OxfordLanguageToken)
+    val oxfordVisionToken: String = envOrFail(Constants.Env.OxfordVisionToken)
+
+    val appInsightsKey: Option[String] = envOrNone(Constants.Env.AppInsightsKey)
+    val modelsDir: Option[String] = envOrNone(Constants.Env.LanguageModelDir)
+
+    private def envOrFail(name: String): String = {
```
c-w: I'd move this out to a utility as the function implementation is unrelated to the code at hand and so just takes up space when trying to scan the code. The brain's L1 cache is pretty limited so we should keep code at the [same level of abstraction](https://dzone.com/articles/the-single-level-of-abstraction-principle).

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
```diff
@@ -0,0 +1,194 @@
...
+  // TODO: move most of these settings to Cassandra
+  private implicit object Settings extends Settings {
+    import scala.util.Properties.envOrNone
+
+    val progressDir: String = envOrFail(Constants.Env.HighlyAvailableProgressDir)
+    val featureServiceHost: String = envOrFail(Constants.Env.FeatureServiceHost)
+    val oxfordLanguageToken: String = envOrFail(Constants.Env.OxfordLanguageToken)
+    val oxfordVisionToken: String = envOrFail(Constants.Env.OxfordVisionToken)
+
+    val appInsightsKey: Option[String] = envOrNone(Constants.Env.AppInsightsKey)
```
c-w: Shouldn't the compiler be able to infer these types? Why do we need to declare them? The type is pretty obvious from the name so declaring it doesn't add much value.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Settings.scala
```diff
@@ -0,0 +1,10 @@
+package com.microsoft.partnercatalyst.fortis.spark
+
+trait Settings {
```
c-w, line 3: Here the trait is a good choice since we'll have multiple implementations. Right now we have an `EnvironmentVariablesSettings` and in the future we'll have a `CasandraSettings`.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/StreamProviderFactory.scala
```diff
@@ -0,0 +1,43 @@
+package com.microsoft.partnercatalyst.fortis.spark
+
+import java.io.File
+
+import com.microsoft.partnercatalyst.fortis.spark.streamfactories.adapters.TadawebAdapter
+import com.microsoft.partnercatalyst.fortis.spark.streamfactories._
```
c-w, line 6: Not a huge fan of wildcard imports since they have [a few negatives](http://seanmonstar.com/post/708954358/import-star-considered-harmful) and don't add much value in the times of IDE-driven auto-imports.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/pipeline/TwitterPipeline.scala
```diff
@@ -0,0 +1,51 @@
...
+import com.microsoft.partnercatalyst.fortis.spark.dto.{Analysis, AnalyzedItem, FortisItem}
+import com.microsoft.partnercatalyst.fortis.spark.streamprovider.{ConnectorConfig, StreamProvider}
+import org.apache.spark.streaming.StreamingContext
+import org.apache.spark.streaming.dstream.DStream
+import twitter4j.{Status => TwitterStatus}
+
+object TwitterPipeline extends Pipeline {
+  override def apply(streamProvider: StreamProvider, streamRegistry: Map[String, List[ConnectorConfig]])
+    (implicit ssc: StreamingContext, transformContext: TransformContext): Option[DStream[FortisItem]] = {
+    import transformContext._
```
c-w: Why not just pass in a transformContext instance and dereference member variables on there? That could would require much less Scala knowledge and would be more easily parsable for folks from a variety of backgrounds.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
```diff
@@ -0,0 +1,194 @@
...
+    // TODO: other computations and save to DB
+    fortisEvents.foreach(
+      _.foreachRDD(
+        _.foreachPartition(throw new NotImplementedException)))
+
+    streamingContext
+  }
+
+  // Main starts here
+  val ssc = StreamingContext.getOrCreate(Settings.progressDir, createStreamingContext)
```
c-w: I take it this is the call that [enables checkpointing](https://spark.apache.org/docs/1.6.1/api/java/org/apache/spark/streaming/StreamingContext.html#getOrCreate(java.lang.String,%20scala.Function0,%20org.apache.hadoop.conf.Configuration,%20boolean)). I was expecting a call to [SparkContext.setCheckpointDir(String)](https://jaceklaskowski.gitbooks.io/mastering-apache-spark/spark-rdd-checkpointing.html) or [StreamingContext.checkpoint(String)](https://spark.apache.org/docs/latest/streaming-programming-guide.html#how-to-configure-checkpointing) which would be more explicit by making it obvious what's going on without having to know the details of the Spark API. See also the [principle of least surprise](http://wiki.c2.com/?PrincipleOfLeastAstonishment).

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
```diff
@@ -0,0 +1,194 @@
...
+            "name" -> System.getenv("TADAWEB_EH_NAME"),
+            "partitionCount" -> System.getenv("TADAWEB_EH_PARTITION_COUNT"),
+            "consumerGroup" -> "$Default"
+          )
+        )
+      )
+    )
+  }
+
+  private object DriverLog {
```
c-w: Why do we need this? It's just used in one place and after the call the JVM is forced to shut down, so we might as well write to `System.err` directly.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
```diff
@@ -0,0 +1,194 @@
...
+            "accessToken" -> System.getenv("TWITTER_ACCESS_TOKEN"),
+            "accessTokenSecret" -> System.getenv("TWITTER_ACCESS_TOKEN_SECRET")
+          )
+        )
+      ),
+      "facebook" -> List(
+        ConnectorConfig(
+          "FacebookPage",
+          Map(
+            "accessToken" -> System.getenv("FACEBOOK_AUTH_TOKEN"),
```
c-w, line 154: Might as well use the getenvOrFail helper created previously since these environment variables are all required if the connector is used. May need changing the map from `Map[String,String]` to `Map[String,()=>String]` so that the values are lazy-loaded.

On second thoughts: shouldn't these all be part of the settings since they'll also have to come from Cassandra going forward?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
```diff
@@ -0,0 +1,194 @@
...
+      new SparkContext(conf),
+      Seconds(Constants.SparkStreamingBatchSizeDefault))
+
+    val streamProvider = StreamProviderFactory.create()
+    val streamRegistry = getStreamRegistry
+
+    // Attach each pipeline (aka code path)
+    // 'fortisEvents' is the stream of analyzed data aggregated (union) from all pipelines
+    val fortisEvents = pipelines.flatMap(
+      pipeline => pipeline(streamProvider, streamRegistry)
```
c-w: Wouldn't `_(streamProvider, streamRegistry)` also work here? If we're already using lots of underscores, might as well be consistent.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
```diff
@@ -0,0 +1,194 @@
...
+    ).reduceOption(_.union(_))
+
+    // TODO: other computations and save to DB
+    fortisEvents.foreach(
+      _.foreachRDD(
+        _.foreachPartition(throw new NotImplementedException)))
```
c-w: Might as well use a print for now.

c-w (approved):
This code is getting too smart for me, see [Kernighan's lever](http://www.linusakesson.net/programming/kernighans-lever/) for reference. I'd appreciate if we'd strive for more simplicity in this project since the project is already complicated enough so that we can't really afford to add extra cognitive overhead, especially given that we're about to onboard lots of new people.

As a more general comment, it would have been ideal to split the refactor and the feature implementation (checkpointing) into separate pull requests or at least separate commits as I had issues finding the code change for the latter in the mass of changes related to the former.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
c-w: Most of our data sources are Spark Receivers. This means that just enabling checkpointing is not enough: we also need to enable the [write-ahead log](https://databricks.com/blog/2015/01/15/improved-driver-fault-tolerance-and-zero-data-loss-in-spark-streaming.html).

image

For more detail, check out the [The Top Five Mistakes Made When Writing Streaming Applications](https://www.youtube.com/watch?v=I7cULsDbSaU) SparkSummit talk around 8:30.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/dto/FortisItem.scala
kevinhartman: The point of this is to allow us to easily drop the original item (T) member once analysis is done are we're ready to pass the analysis into the downstream code path. I would have used a case class here, but IIRC, you cannot extend a case class with another case class.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/StreamProviderFactory.scala
kevinhartman: In general, I agree. In this case, I don't see the harm. We control the contents of the stream factories package, and we'll need to instantiate all of those factories within this class.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/StreamProviderFactory.scala
kevinhartman: For Settings specifically, it's intended to be somewhat of a global variable. I think that makes it a great candidate for implicit. We can treat it as such without actually coupling our code to a singleton.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
kevinhartman: I think you're right that it should be able to infer it. I'd generated the body of this implementation. I'll see if I can remove the explicit types here.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
kevinhartman: It's just a quick and dirty helper function, with an implementation very specific to the driver program. It's not intended to be used by other pieces of code. Having it here prevents others from taking dependency on it.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
kevinhartman: IMO, the interplay between the syntactic sugars for function application and ordered argument reference is pretty unreadable. I'm sure you probably agree :p

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
kevinhartman: I agree with the principle of least surprise in general, though `getOrCreate` is the method used in the Spark Streaming Programming Guide's [How to configure checkpointing](https://spark.apache.org/docs/latest/streaming-programming-guide.html#how-to-configure-checkpointing) section. This should be a perfectly expected choice.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
kevinhartman: Connector config stuff won't be on the Settings object, but yeah the reason I didn't use getEnvOrFail here is because this code should go away. It's essentially stubbing for now.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
kevinhartman: I'd agree that System.err would be a better choice, however we need the logs to go up to AppInsights, right? Will that still work without using log4j? Happy to remove this if so.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
kevinhartman: I did see that. I'm not sure that our requirements make sense with checkpointing at all TBH. I will explain my thoughts on that in standup.

I can enable WAL as well for this PR, but I haven't looked into the considerations needed with doing that yet.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/dto/FortisItem.scala
c-w: Now that we have a clearer idea of the downstream Cassandra schema, we can soon get rid of `originalItem` in favor of more general fields like `title` and `body`. I'll make a follow-up PR once this one is merged.

PR #20, opened 2017-06-20
Improve serialization performance

### LINKED ISSUE #15: Spark Pipeline - Add Kryo as our serialization interface
We should avoid using the default serialization interface that's shipped with spark due to performance issues. Kryo seems to be the industry standard serialization solution across the Java community.

*What does complete mean*
Add the kryo jar as a dependency in our build, and configure the serialization interface with `SparkContext`. This work also includes the unit testing and performance tuning with the library.

c-w:
This PR switches our Spark setup to using the Kryo serializer for common DTO classes.

Unfortunately we can't enforce the use of the Kryo serializer for all classes (via the setting `spark.kryo.registrationRequired`) because some classes that we depend upon are non-public so we can't easily get a handle to them to register them in Kryo (e.g. TwitterReceiver).

Resolves #15

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -31,6 +31,7 @@ libraryDependencies ++= Seq(
   "org.apache.bahir" %% "spark-streaming-twitter" % "2.1.0",
   "org.apache.commons" % "commons-collections4" % "4.1",
   "com.microsoft.azure" %% "spark-streaming-eventhubs" % "2.0.5",
+  "com.esotericsoftware.kryo" % "kryo" % "2.24.0",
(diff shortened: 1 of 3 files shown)


kevinhartman (approved):
LGTM. Perhaps we can register some internal classes using reflection in a later PR?

c-w:
I did some research and didn't find anyone doing this. From my understanding, it's usually good enough to use Kryo for the commonly serialized data types (i.e., the DTOs) and keep using the Java serializer for less common types (like Receivers).

PR #22, opened 2017-06-21
Rationalize item processing pipelines

c-w:
We currently have a bunch of copy/paste code in the pipelines package which leads to poor maintainability because we need to make changes for new feature extractors in multiple places and because we can't guarantee uniform processing across all pipelines.

This pull request removes the copy/paste code and rationalizes the processing pipelines by introducing the TextPipeline class which is used by the Twitter, Facebook, Radio and Tadaweb pipelines.

### DIFF
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/dto/AnalyzedItem.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/dto/AnalyzedItem.scala
@@ -1,11 +1,12 @@
 package com.microsoft.partnercatalyst.fortis.spark.dto
 
-case class AnalyzedItem[T](
-  originalItem: T,
+case class AnalyzedItem(
+  body: String,
+  title: String,
   source: String,
   sharedLocations: List[Location] = List(),
   analysis: Analysis
-) extends FortisItem
+)
 
 case class Analysis(
   language: Option[String] = None,
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/dto/FortisItem.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/dto/FortisItem.scala
@@ -1,7 +1,0 @@
-package com.microsoft.partnercatalyst.fortis.spark.dto
-
-trait FortisItem {
-  def source: String
-  def sharedLocations: List[Location]
-  def analysis: Analysis
-}
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/pipeline/FacebookPipeline.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/pipeline/FacebookPipeline.scala
@@ -1,52 +1,31 @@
 package com.microsoft.partnercatalyst.fortis.spark.pipeline
 
 import com.github.catalystcode.fortis.spark.streaming.facebook.dto.FacebookPost
-import com.microsoft.partnercatalyst.fortis.spark.dto.{Analysis, AnalyzedItem, FortisItem}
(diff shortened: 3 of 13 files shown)


kevinhartman:
Happy to see the duplication reduced. The issue I have with this approach however is that for the broadcasting and shared state stuff I'm working on now, it'd be ideal to inject a `transform` at the very start of every pipeline, and to ideally have the logic within it be the same.

I've adapted the concepts from this PR a bit into a proof of concept configuration with inheritance. This will allow me to inject the code that I need for state updates in the `transform` [call in the base class.](https://github.com/CatalystCode/project-fortis-spark/compare/CatalystCode:26202d7...kevinhartman:4ee2cf7#diff-200f5e9898092a51f7aa5acbcedf5f9dR29)

I've written the code for the Tadaweb pipeline to illustrate how this would come together. Note that the TransformContext is not used directly by `TadawebPipeline` on purpose. This is because it will be removed when the state update stuff comes in.

Aside, I think your initial thinking of keeping the original item around in the analyzed item is the right approach. It seems more future proof to me to defer the actual analysis of a unique struct to its pipeline rather than trying to stuff everything into a generic struct. The approach below abstracts the pipeline steps/order away from the individual pipelines but allows them to implement each how they like in the context of their own types.

What do you think?
https://github.com/CatalystCode/project-fortis-spark/compare/CatalystCode:26202d7...kevinhartman:4ee2cf7

c-w:
As per my [comment](https://github.com/CatalystCode/project-fortis-spark/pull/11#discussion_r123098238) on your last pull request, we don't need the originalItem anymore now that we have a better idea of the Cassandra schema. There is a massive benefit of getting rid of originalItem, namely being able to treat all text sources the same no matter where the text comes. The downstream processing really shouldn't have to care about whether it's a Tweet or a Facebook post just to extract keywords or sentiment. Using a single structure also reduces the code complexity which is a pro. Also note that at the end of the processing all data will have to fit within a single structure anyways so that it can be persisted to Cassandra.

If we need to inject an additional transform, that's a separate concern which can be tackled in multiple ways. For example, if the transform truly needs to be applied to every stream, just do it in the StreamProvider's buildStream: buildStream should return a downstream usable stream so that transforms don't have to care about any book-keeping. Alternatively, setup code can also live in ProjectFortis so that we'd have a section in there for setup, then processing and then sinks.

kevinhartman:
It's not a question of knowing the DB schema, but rather knowing the formatting and analysis requirements of the incoming items. If an item has, for example, 2 separate body fields which need keyword analysis, we should be able to handle that. If we need to **override** sentiment analysis because the source item already has a better one, we need to support that. If we want to skip keyword extraction for the `title` of an item, we should support that. Imagine a case where we want to include sentiment analysis on the title but not the body.

We lose flexibility with this approach.

c-w:
We currently don't have these requirements and don't plan to add any other data sources. As such, I'd be hesitant to add complexity for potential future needs instead of focusing on what we need now and refactor later if necessary. Going from this type of function-based code to an object-based code is an easy refactor if it becomes necessary (just turn TextPipeline into an object with methods for each step and instantiate a custom subclass in the caller to do the processing). Note that if you really need to do something very custom, you can just not use the TextPipeline (see InstagramPipeline for an example).

kevinhartman:
We *already* have one of these requirements: we should not perform our own sentiment analysis if Tadaweb has already provided it. We're also adding Bing in the next week or so :p 

This LGTM to get us going. I'll sign off after reviewing (tomorrow morning).

c-w:
Seems odd to me that we'd trust a third party more than our own analysis, but sure, if that's the case, it's trivial to fix, see 14c9e3c.

As for Bing: it'll likely be the same processing as Facebook/Twitter.

kevinhartman (approved):
LGTM

PR #23, opened 2017-06-21
Add registration-order disclaimer

c-w:


jcjimenez (approved):
LGTM - it would also be worth considering using the `#register` method that takes an `int` parameter and use assign ranges like 1000 for fortis, 1100 for twitter, etc.

PR #24, opened 2017-06-21
Deduping Analysis tag element values.

jcjimenez:
Converting Analysis tag list fields to sets.


jcjimenez:
Closing to test further.

c-w:
I wonder if we want to keep the tags as lists instead of sets. E.g. is it important for us to know that Isis was mentioned twice in a post? This is a question for @erikschlegel.

PR #25, opened 2017-06-22
Add Bing connector

### LINKED ISSUE #5: Integrate Bing Spark Streaming Connector to Fortis
The `DemoFortis` app should provide a sample transformation on how to integrate the unionized `DStream` with events from Bing Streaming Connector

c-w:
Resolves #5


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
```diff
@@ -128,6 +129,16 @@ object ProjectFortis extends App {
           )
         )
       ),
+      "bing" -> List(
+        ConnectorConfig(
+          "BingPage",
+          Map(
+            "accessToken" -> System.getenv("BING_ACCESS_TOKEN"),
+            "searchInstanceId" -> System.getenv("BING_SEARCH_INSTANCE_ID"),
+            "keywords" -> "isis|attack"
```
jcjimenez, line 138: Not a deal-breaker, but may I suggest `Option(System.getenv("BING_SEARCH_KEYWORDS")).getOrElse("isis|attack")` instead of the raw term?

jcjimenez (approved):
LGTM with minor observation

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
c-w: Good comment. These values are just here for testing purposes for now so they're fine to be hard-coded (see also all the other hard-coded values in this section). We'll get rid of them soon and in production, we'll populate the configs from Cassandra which is what Kevin is about to implement.

jcjimenez: Perfect!

PR #27, opened 2017-06-22
Add more fields required for Cassandra schema

c-w:


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/pipeline/BingPipeline.scala
```diff
@@ -15,9 +17,11 @@ object BingPipeline extends Pipeline {
 
   private def convertToSchema(stream: DStream[BingPost], transformContext: TransformContext): DStream[AnalyzedItem] = {
     stream.map(post => AnalyzedItem(
+      createdAtEpoch = now.getEpochSecond,
```
jcjimenez, line 20: Question: Do these incoming posts have a date of their own? If so, do we prefer to use our own capture time as opposed to the timestamp in the post?

jcjimenez (approved):
LGTM with minor question.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/pipeline/BingPipeline.scala
c-w: They have a crawl-date but not a post-date ([BingPost.scala](https://github.com/CatalystCode/streaming-bing/blob/d1fdc2f4d5f69b052434c25429cc7ddc233eab38/src/main/scala/com/github/catalystcode/fortis/spark/streaming/bing/dto/BingPost.scala#L3-L8)). Not sure if that's something that would be more useful to us than a timestamp for which we definitely know what it means.

PR #29, opened 2017-06-26
Broadcast and Configuration Update Foundation

kevinhartman:
<b>Changes</b>
* Decouples the analysis stream processing definition (`DStream`) from corresponding pipeline analysis implementations which:
  * Provides a single place ahead of all processing for the transform block that is required in order to perform live updates of configuration and broadcasted data (and is run on the driver). Config updates must occur here, along with broadcasts of those updates where applicable.
  * Draws on pull request #22 by reusing common analysis methods introduced there but with added support for override; concrete `Analyzer`s (one impl per pipeline) are responsible for analyzing their corresponding event types using the tools provided to them by the pipeline, but can choose to use default implementations where applicable. The pipeline (stages and their ordering) is completely hidden from `Analyzer` implementations, which allows us to rewrite and augment it without modifying them (the only file defining the analysis pipeline is `Pipeline.scala`).
  * Removes the need for TransformContext (no need to pass all transform tools into multiple files), whose members were previously being serialized to each task regardless of whether or not the task used them.
* Adds the configuration/broadcast update entrypoint mentioned above as a foundation for a future config update changeset.
* Adds a `ConfigurationManager` interface for DB access of config settings. Currently includes stream config only, but will include methods for accessing transform tool settings as well.
* Moves `language` parameter from transform tool methods into their constructors. In this way, `Analyzers` do not need to know the languages of their events in order to use transformation tools. This also provides a cleaner interface for future implementations which may be language agnostic. Note that object construction occurs within tasks on workers. Only the model provider (path and local path) is serialized to workers (and `LocationsExtractorFactory` in the case of locations).
* Adds interfaces for Fortis model schema to enable custom implementations. The Analysis layer uses a custom implementation in order to keep raw event information around until after analysis is complete.
* Renames `AnalyzedItem` to `FortisEvent`.
* Fixes a few readonly fields denoted as `volatile` which were likely intended to be `transient`.


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
```diff
@@ -0,0 +1,100 @@
...
+        val keywords = analyzer.extractKeywords(event.details, keywordExtractor)
+        event.copy(analysis = event.analysis.copy(keywords = keywords))
+      }
+
+      def hasKeywords(analysis: Analysis): Boolean = {
+        analysis.keywords.nonEmpty
+      }
+
+      def addEntities(event: ExtendedFortisEvent[T]): ExtendedFortisEvent[T] = {
+        val entities = analyzer.extractEntities(event.details, new PeopleRecognizer(modelsProvider, event.analysis.language))
```
c-w, line 74: It's expensive to instantiate this object (needs to download multi-megabyte model files and then load them into memory) so we should avoid doing so. The previous version with one static recognizer that does not take the language in the constructor avoids this issue.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/ExtendedFortisEvent.scala
```diff
@@ -0,0 +1,21 @@
...
+
+case class ExtendedDetails[T](
+  id: UUID,
+  createdAtEpoch: Long,
+  body: String,
+  title: String,
+  publisher: String,
+  sourceUrl: String,
+  sharedLocations: List[Location] = List(),
+  original: T
```
c-w, line 20: Why do we need `original: T` here? As per [previous discussion](https://github.com/CatalystCode/project-fortis-spark/pull/22#issuecomment-310201771), we don't need to keep around the original item. Not keeping the item around (a) ensures that we don't encounter any surprises in the last stage of the pipeline when we try to persist all the disparate schemas to Cassandra and (b) minimizes the amount of data we have to unnecessarily serialize for every stage of the Spark pipeline.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
```diff
@@ -0,0 +1,100 @@
...
+      }
+
+      // Configure analysis pipeline
+      rdd
+        .map(convertToSchema)
+        .map(addLanguage)
+        .filter(item => isLanguageSupported(item.analysis))
+        .map(addKeywords)
+        .filter(item => hasKeywords(item.analysis))
+        .map(item => addLocations(addSentiments(addEntities(item))))
```
c-w, line 97: The previous code spread out the operations over multiple workers as each of these operations (`addLocations`, `addSentiments` and `addEntities`) is expensive, adds value on its own and has no dependency on the previous operations.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
```diff
@@ -0,0 +1,100 @@
...
+    val configs = configurationManager.fetchStreamConfiguration(name)
+    val sourceStream = streamProvider.buildStream[T](ssc, configs)
+
+    sourceStream.map(_.transform(rdd => {
+
+      // TODO: populate these local variables with values from global update/broadcast wrappers,
+      // which will use the ConfigurationManager (cassandra) to update themselves and (where applicable)
+      // broadcast changes across the Spark cluster.
+
+      val geofence = Geofence(north = 49.6185146245, west = -124.9578052195, south = 46.8691952854, east = -121.0945042053)
```
c-w, line 33: I don't understand why we're instantiating these objects here instead of creating them once, wrapping them in the TransformContext that we previously used and then broadcasting the TransformContext to the workers ([example code](https://stackoverflow.com/a/36660199/3817588)). To me that seems as though it should provide all the functionality we need.

c-w (changes requested):
Like some previous pull requests, this change set combines a large refactor with a small addition of functionality. Ideally, refactoring and functional changes would be submitted in separate code reviews as it'd make the changes easier to review.

Additionally, I'm a bit concerned that we need to refactor large parts of the codebase to add minimal new functionality. Maybe it would be good to discuss implementation plans on the tasks before starting work to avoid having to do these large refactors in favor of making more surgical changes to implement new functionality? E.g. there may have been some ways to implement broadcast variables with more minimal changes. Large changes are painful for the team (merge conflicts, unstable core paradigm) and slow down our feature velocity.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
kevinhartman: That's a great question, with a non-obvious answer that has to do with the serialization nuances of checkpointing. The reason for doing this is because we need to poll Cassandra here and possibly update the transform tooling (with new settings etc.). This has to be a polling-based approach from within the DStream definition itself since when Spark comes back up from a checkpoint, we will no longer be able to affect any of its closures. The closure of the `transform` function here is serialized to the checkpoint, so when Spark comes back up, the checkpoint must know how to update itself.

Instead of `new`ing up instances of the transform tools here directly, we will call into wrapper objects which will recycle or regenerate (and rebroadcast where applicable) the instances of these tools depending on if there is a configuration change. This will happen once per batch on the driver program (so maybe once every 5 to 10 seconds).

kevinhartman: For an example of what the broadcast wrappers will look like (minus Cassandra updates), check out the official broadcasts with checkpointing example code: https://github.com/apache/spark/blob/master/examples/src/main/scala/org/apache/spark/examples/streaming/RecoverableNetworkWordCount.scala#L35

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
kevinhartman: In what way does the previous code avoid a data dependency? Each stage in the existing code takes an `AnalyzedItem` and outputs a new instance of an `AnalyzedItem`. How can Spark do any of those stages in parallel?

Aside, I'm not convinced that scaling the analysis of a single event across multiple workers is beneficial. Our scaling comes from the ability to consume many events in parallel. Do we really expect to have enough resources to assign N > 1 nodes per incoming event? I think our best bet for optimization of these three calls in general is to use a single `map` call and write our own multi-threaded logic there to execute on a single worker. This implementation can easily accommodate that as well.

c-w: Having the transforms as a single `map` call means that for sure they will run as a single computation unit on a single worker. Having multiple `map` calls means that the Spark DAG scheduler can optimize the computation across the cluster which in the worst case will run them in sequence on the same node. Having multiple `map` calls is thus more flexible (plus, it makes the code more explicit in the processing steps that we do).

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
c-w: Okay. In that case, can we create a TransformContext from the broadcast variables in the `forEachRDD` loop and pass that to the pipelines? No need to couple the broadcasting to the transform logic (plus a much smaller required change than the current approach).

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/ExtendedFortisEvent.scala
kevinhartman: As I'd commented on that discussion, there are a lot of issues with assuming we can converge every unique data model into a single unanalyzed representation.

I'd mentioned that we needed to *override* sentiment analysis for Tadaweb, and to support this, the old implementation was modified to *always* override if an analysis was already present. That's also not necessarily correct, since perhaps another pipeline will need to only augment the analysis.

The current `TextPipeline` is already getting hairy. Concrete `Analyzer`s that can handle analysis how they like allows for cleaner common code and more flexibility/extensibility. Fortis will one day be maintained by the community, and so we should design with extensibility in mind.

Serializing a few extra properties should not be an issue, since this will only happen once per event, given that our pipeline is largely sequential. See my comment above on line `97` of `Pipeline.scala`.

c-w: Not sure what you mean by hairy given that it's 90 lines of code with about 50 functional lines. There are many potential eventualities that we can't account for. As such, it's usually wise to defer generalization until the requirement actually exists.

My main concern here is that we spent all this time shuffling code around instead of getting in and making the minimum necessary change to get our required features built out. We're up against some pretty tight deadlines so it'd be great to avoid big changes in the future and implement what we need with the minimal amount of investment... here, unfortunately the time is already spent and I understand you're convinced that this is the right way forward, so there is little value in continuing to discuss this.

c-w (approved):

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
kevinhartman: I believe the DAG actually optimizes by *combining* the maps calls, only. Imagine the simple case of an `RDD[String]`. If you define a series of maps on it, each map call has a data dependency on the previous. The point of Spark is to partition the data across nodes and then bring transformations to the data. I doubt the DAG will create copies of the same element on each node, distribute different closures, and then somehow (how can it?) merge the elements back together.

I can certainly move these into their own map calls for readability (since I'm almost certain Spark will merge them anyway :) ) TBH, I only just realized that this line is what you were concerned about (forgot I'd combined these :p).

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
kevinhartman: This needs to be a transform call since we want this to happen at the beginning of the pipeline. To do that in the current implementation, we'd need to put a `transform` call in every single `*Pipeline.scala` since they use transform utils.

The reality is that the current pipeline architecture (the one I'd started and you'd finished) was not designed with checkpointing in mind. This PR significantly strengthens our architecture around serialization concerns for checkpointing and task creation. Smaller change isn't necessarily best when considering architecture. This will not happen typically—we did not understand the nuances of Spark and serialization, which should have been a pre-requisite for our initial design.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/ExtendedFortisEvent.scala
kevinhartman: Again, the requirement already exists, even if we haven't hit the eventuality that show-stopping conflicts between pipeline requirements can occur with the current approach. I wouldn't waste my own time if I didn't see a clear issue.

I completely agree that we've wasted time, but the time wasn't wasted through this refactor, but rather at the point when we'd jumped into an initial design without research and discussion. That's on all of us as a team. FWIW, this PR makes extensive use of the implementation patterns (if not verbatim copies) introduced in #22.

The purpose of this PR's comment section is to discuss its implementation and to communicate design decisions / transfer understanding. If it has not yet served that purpose, I encourage you to add more discussion points. I'm happy to explain any choices I've made here, and to ideally convince you that this is a good way forward as well.

c-w: Don't get me wrong; this approach is likely better than what there was before. I just question whether now was the right time to make these changes as opposed to just implementing what we needed, getting on with feature work and cleaning things up later if required. Given that the change is here, it's fine. That's why I signed off. It's just that this is the second huge pull request in two weeks, so going forward I'm wondering if we can implement functionality with fewer, smaller and more surgical changes.

I disagree that we wasted time with the initial approach as getting something up and running on which to iterate is always worth something. Now it's just about rearranging the blocks :)

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/AnalysisDefaults.scala
```diff
@@ -0,0 +1,62 @@
...
+import com.microsoft.partnercatalyst.fortis.spark.transforms.topic.KeywordExtractor
+
+/**
+  * Provides default analysis method implementations for a concrete [[Analyzer]].
+  *
+  * By implementing these traits, the concrete [[Analyzer]] explicitly enables the default implementations that are
+  * applicable to it. If a default is not enabled, the compiler will enforce that a custom implementation is provided.
+  *
+  */
+private[analyzer] object AnalysisDefaults {
```
erikschlegel, line 17: I'm sure you considered this but why cant we define these methods in a single base class as opposed to defining so many individual traits.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
```diff
@@ -103,88 +86,90 @@ object ProjectFortis extends App {
...
-            "tag" -> "rose"
+  private object DummyConfigurationManager extends ConfigurationManager {
+    override def fetchStreamConfiguration(pipeline: String): List[ConnectorConfig] = {
+      // The key is the name of the pipeline and the value is a list of connector configs whose streams should comprise it.
+      Map[String, List[ConnectorConfig]](
+        "instagram" -> List(
+          ConnectorConfig(
+            "InstagramTag",
+            Map(
+              "authToken" -> System.getenv("INSTAGRAM_AUTH_TOKEN"),
```
erikschlegel, line 101: Can we move these out of the environment and source these from the `streams` table.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/Analyzer.scala
```diff
@@ -0,0 +1,20 @@
...
+import com.microsoft.partnercatalyst.fortis.spark.transforms.language.LanguageDetector
+import com.microsoft.partnercatalyst.fortis.spark.transforms.locations.LocationsExtractor
+import com.microsoft.partnercatalyst.fortis.spark.transforms.people.PeopleRecognizer
+import com.microsoft.partnercatalyst.fortis.spark.transforms.sentiment.SentimentDetector
+import com.microsoft.partnercatalyst.fortis.spark.transforms.topic.KeywordExtractor
+
+trait Analyzer[T] {
+  type LocationFetcher = (Double, Double) => Iterable[Location]
+
+  def toSchema(item: T, locationFetcher: LocationFetcher, imageAnalyzer: ImageAnalyzer): ExtendedDetails[T]
```
erikschlegel, line 14: add a definition for gender as well.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
```diff
@@ -0,0 +1,100 @@
...
+import com.microsoft.partnercatalyst.fortis.spark.transforms.topic.KeywordExtractor
+import org.apache.spark.streaming.StreamingContext
+import org.apache.spark.streaming.dstream.DStream
+
+object Pipeline {
+  def apply[T: TypeTag](name: String, analyzer: Analyzer[T], ssc: StreamingContext, streamProvider: StreamProvider, configurationManager: ConfigurationManager): Option[DStream[FortisEvent]] = {
+    val configs = configurationManager.fetchStreamConfiguration(name)
+    val sourceStream = streamProvider.buildStream[T](ssc, configs)
+
+    sourceStream.map(_.transform(rdd => {
```
erikschlegel, line 27: I'd like to better understand how things like topic keywords and the geofence will be synchronized from Cassandra into this class. I see your TODO below, so perhaps we should schedule a targeted discussion to walk through your implementation strategy.

erikschlegel (approved):
LGTM. Left some general comments but should be good to merge. We should have a general design discussion with yourself and @c-w to chat about your implementation strategy re: broadcast vars.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
kevinhartman: Yes definitely. That will be a part of the broadcast and state update task that I should have a PR for this week.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/AnalysisDefaults.scala
kevinhartman: I'd considered that, but I believe default analysis methods make more sense as an opt-in feature since it's possible that they aren't applicable for every pipeline. It seems like a code smell to me for N.A. methods to be lurking in a base class, ready for someone to forget they're there.

Really this is just syntactic sugar for calling a shared helper method explicitly for each concrete method implementation. If an Analyzer implements a particular default trait, then the corresponding analysis method implementation will be present in the base class (and defaulted to if the concrete Analyzer does not override it, or callable with super() if it does). 

In other words, a default analysis implementation will be available in the base class hierarchy of a concrete `Analyzer` if and only if the concrete Analyzer explicitly denotes that a default implementation is applicable to it by implementing the trait.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/Analyzer.scala
kevinhartman: Have I omitted this from the original implementation, or is this something that we still need to bring online with an additional changeset? I'd seen gender stuff in an earlier PR, but analysis wasn't yet being performed IIRC.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
kevinhartman: That'd be fine.

In short, for each batch, we'll check Cassandra for configuration changes (every 5 to 10 seconds or so, just once on the driver). Let's say the keyword whitelist is updated. We will fetch that data from Cassandra (still within the `transform` function, and hence still on the driver), and then broadcast it (also from within `transform`). The updated data will be available for use by the `rdd` pipeline stages below through the broadcast handle from the enclosing `transform` scope. 

Note that we need to bring the data from Cassandra to the driver before broadcasting since it's only possible to broadcast data out from the driver to workers, and not from worker to workers. This isn't a drawback, but might be useful for your understanding.

In terms of efficiency of updates to already broadcasted information, unfortunately there's no concept of transferring just the delta. In practice, this should be fine since we will only rebroadcast when the user has updated terms. Also note that broadcasting itself doesn't incur transfer penalty. Each node has a cache of broadcasted data that, on miss, will fetch from the driver.

PR #32, opened 2017-06-27
Remove comment that causes warning on sbt assembly

c-w:


kevinhartman (approved):
LGTM. Interesting. What's the warning here?

c-w:
I didn't grab the console output, but it was something about something not being a valid target. Anyways, the comment doesn't add much value on top of the method signature, so good riddance to it. Less green is always a good thing.

PR #36, opened 2017-06-27
Implement custom event pipeline

### LINKED ISSUE #30: Add pipeline for custom events
Fortis-v1 has a feature where users can upload custom events and have them analyzed via the Fortis pipeline. See [publishEvents](https://github.com/CatalystCode/project-fortis-services/blob/1c2e6ad6939d0c3e5eed2e6e9622460ae714103c/src/schemas/MessageSchema.js#L19-L46) for reference.

If we want to support this feature in Fortis-v2, we need to add a [pipeline](https://github.com/CatalystCode/project-fortis-spark/tree/257576dcaa2b72a13109c30c3edb5cb0986a139c/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/pipeline) via which to ingest these custom events.

One possible implementation of this is to use [EventHub](https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-what-is-event-hubs) with the Spark pipeline acting as a consumer and the resolver that implements publishEvents acting as a producer.

We already have code to interface with EventHub from Spark ([EventHubStreamFactory](https://github.com/CatalystCode/project-fortis-spark/blob/257576dcaa2b72a13109c30c3edb5cb0986a139c/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/streamfactories/EventHubStreamFactory.scala#L13)) so it shouldn't take more than 1-1.5 days to build all the required pieces:

(1) Define the schema for the custom events
(2) Define a name for the custom events EventHub
(3) Add a new pipeline that processes events from EventHub
(4) Set up a dummy EventHub in our Azure subscription for testing

c-w:
[Test EventHub](https://ms.portal.azure.com/#resource/subscriptions/04f7ec88-8e28-41ed-8537-5e17766001f5/resourceGroups/fortis-resource-group/providers/Microsoft.EventHub/namespaces/fortiscustomevents/overview) to play with this.

Resolves #30

### DIFF
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
@@ -11,6 +11,7 @@ import org.apache.spark.streaming.{Seconds, StreamingContext}
 import org.apache.spark.{SparkConf, SparkContext}
 
 import scala.reflect.runtime.universe.TypeTag
+import scala.util.Properties.envOrElse
 
 object ProjectFortis extends App {
 
@@ -71,6 +72,7 @@ object ProjectFortis extends App {
       pipeline("facebook", new FacebookAnalyzer),
       pipeline("instagram", new InstagramAnalyzer),
       pipeline("tadaweb", new TadawebAnalyzer),
+      pipeline("customevents", new CustomEventAnalyzer),
       pipeline("bing", new BingAnalyzer),
       pipeline("radio", new RadioAnalyzer)
     ).flatten.reduceOption(_.union(_))
@@ -167,6 +169,19 @@ object ProjectFortis extends App {
               "name" -> System.getenv("TADAWEB_EH_NAME"),
               "partitionCount" -> System.getenv("TADAWEB_EH_PARTITION_COUNT"),
               "consumerGroup" -> "$Default"
+            )
+          )
+        ),
+        "customevents" -> List(
+          ConnectorConfig(
+            "CustomEvents",
+            Map (
+              "policyName" -> envOrElse("CUSTOMEVENTS_EH_POLICY_NAME", "project-fortis-spark"),
+              "policyKey" -> System.getenv("CUSTOMEVENTS_EH_POLICY_KEY"),
+              "namespace" -> envOrElse("CUSTOMEVENTS_EH_NAMESPACE", "fortiscustomevents"),
+              "name" -> envOrElse("CUSTOMEVENTS_EH_NAME", "customevents"),
+              "partitionCount" -> envOrElse("CUSTOMEVENTS_EH_PARTITION_COUNT", "1"),
+              "consumerGroup" -> envOrElse("CUSTOMEVENTS_EH_CONSUMER_GROUP", "$Default")
             )
           )
         )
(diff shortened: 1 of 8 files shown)


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
```diff
@@ -168,6 +170,19 @@ object ProjectFortis extends App {
               "consumerGroup" -> "$Default"
             )
           )
+        ),
+        "customevents" -> List(
+          ConnectorConfig(
+            "CustomEvents",
+            Map (
+              "policyName" -> envOrElse("CUSTOMEVENTS_EH_POLICY_NAME", "project-fortis-spark"),
```
erikschlegel, line 179: from an environment standpoint Im keeping things consistent with GQL, and setting only three vars

```--from-literal=PUBLISH_EVENTS_EVENTHUB_CONNECTION_STRING="${eh_conn_str}" --from-literal=PUBLISH_EVENTS_EVENTHUB_PATH="${eh_path}" --from-literal=PUBLISH_EVENTS_EVENTHUB_PARTITION="${eh_consumer_group}"```

File: src/main/scala/com/microsoft/partnercatalyst/fortis/spark/StreamProviderFactory.scala [hunk 2]
```diff
@@ -42,6 +43,12 @@ object StreamProviderFactory {
             new File(settings.progressDir, Constants.EventHubProgressDir).getPath)
         )
       )
+      .withFactories(
+        List(
```
erikschlegel, line 47: this is going to mean that @Smarker will  have to add a record into the `streams` table upon a called to `CreateOrReplaceSite`, as long as the eventhub environment is properly setup.

erikschlegel (dismissed):
Thanks! please see my comment below re: consolidating the EH related environment so it's inline with graphql.

kevinhartman (approved):
LGTM

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
c-w: Unfortunately the Spark connector for EventHubs [doesn't support the connection string](https://github.com/hdinsight/spark-eventhubs/blob/master/docs/direct_stream.md#how-we-use-it). I'll update the GQL library instead.

c-w: Oh wow, the nodejs EventHubs library [only supports the connection-string](https://github.com/Azure/azure-event-hubs-node/blob/1084724891a92ef6d1a5f1732a52d24ce871dee9/send_receive/lib/config.d.ts#L7-L8). Looks like we'll have to provide both unfortunately.

erikschlegel: Spark can just parse the connection string which is just delimited `;`

erikschlegel: Azure only gives me the connection string so it's going to need to get parsed one way or another 😞

c-w: Can't you fetch the shared access policy and get the details from there? It's all exposed in the Azure portal ([deep link](https://ms.portal.azure.com/#resource/subscriptions/04f7ec88-8e28-41ed-8537-5e17766001f5/resourceGroups/fortis-resource-group/providers/Microsoft.EventHub/namespaces/fortiscustomevents/saskey)).

image

c-w: You should be able to get the auth keys via the EventHubs API: https://docs.microsoft.com/en-us/rest/api/eventhub/eventhubs#EventHubs_RegenerateKeys

PR #37, opened 2017-06-27
Implement token blacklist filtering

### LINKED ISSUE #34: Add terms blacklist
We need to add a feature to our spark pipeline that allows us to drop any events that contain certain keywords.

The implementation can be as simple as creating a [Pipeline filter](https://github.com/CatalystCode/project-fortis-spark/blob/83e871037dbfb47296b5b24d549f8b0c4ddc5321/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala#L92) that uses [Tokenizer](https://github.com/CatalystCode/project-fortis-spark/blob/83e871037dbfb47296b5b24d549f8b0c4ddc5321/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/nlp/Tokenizer.scala#L6) to split the [title/body](https://github.com/CatalystCode/project-fortis-spark/blob/83e871037dbfb47296b5b24d549f8b0c4ddc5321/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/ExtendedFortisEvent.scala#L15-L16) of the Fortis events into words and then checks each word against a blacklist set.

Like the other [shared settings](https://github.com/CatalystCode/project-fortis-spark/blob/83e871037dbfb47296b5b24d549f8b0c4ddc5321/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala#L45), the keyword blacklist should be populated from the [fortis.blacklist table](https://github.com/CatalystCode/fortisdeploy/blob/119d3d847826b2cbd45791b3cf41c2ce1d60144d/ops/storage-ddls/cassandra-setup.cql#L14-L18) in Cassandra and published to the workers as a broadcast variable.

As mentioned in [project-fortis-spark#17](https://github.com/CatalystCode/project-fortis-spark/issues/17), this is required to implement the `{modify,remove}Blacklist` mutation in the `SettingsSchema`.

c-w:
Built without using the keyword extractor so that we can offer more predictable behavior to the user: events will only get filtered out if tokens match exactly what the user specified.

Resolves #34

### DIFF
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
@@ -6,7 +6,7 @@ import com.microsoft.partnercatalyst.fortis.spark.ProjectFortis.Settings
 import scala.reflect.runtime.universe.TypeTag
 import com.microsoft.partnercatalyst.fortis.spark.analyzer.{Analyzer, ExtendedFortisEvent}
 import com.microsoft.partnercatalyst.fortis.spark.dba.ConfigurationManager
-import com.microsoft.partnercatalyst.fortis.spark.dto.{Analysis, Details, FortisEvent}
+import com.microsoft.partnercatalyst.fortis.spark.dto.{Analysis, FortisEvent}
 import com.microsoft.partnercatalyst.fortis.spark.streamprovider.StreamProvider
 import com.microsoft.partnercatalyst.fortis.spark.transforms.ZipModelsProvider
 import com.microsoft.partnercatalyst.fortis.spark.transforms.image.{ImageAnalysisAuth, ImageAnalyzer}
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
@@ -71,8 +71,8 @@ object Pipeline {
         analysis.keywords.nonEmpty
       }
 
-      def hasBlacklistedTerms(details: Details): Boolean = {
-        blacklist.matches(details.body) || blacklist.matches(details.title)
+      def hasBlacklistedTerms(event: ExtendedFortisEvent[T]): Boolean = {
+        analyzer.hasBlacklistedTerms(event.details, blacklist)
       }
 
       def addEntities(event: ExtendedFortisEvent[T]): ExtendedFortisEvent[T] = {
@@ -95,7 +95,7 @@ object Pipeline {
       // Configure analysis pipeline
       rdd
         .map(convertToSchema)
-        .filter(item => !hasBlacklistedTerms(item.details))
+        .filter(item => !hasBlacklistedTerms(item))
         .map(addLanguage)
         .filter(item => isLanguageSupported(item.analysis))
         .map(addKeywords)
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
@@ -16,7 +16,7 @@ import com.microsoft.partnercatalyst.fortis.spark.transforms.locations.client.Fe
 import com.microsoft.partnercatalyst.fortis.spark.transforms.nlp.Tokenizer
 import com.microsoft.partnercatalyst.fortis.spark.transforms.people.PeopleRecognizer
 import com.microsoft.partnercatalyst.fortis.spark.transforms.sentiment.{SentimentDetector, SentimentDetectorAuth}
-import com.microsoft.partnercatalyst.fortis.spark.transforms.topic.KeywordExtractor
+import com.microsoft.partnercatalyst.fortis.spark.transforms.topic.{Blacklist, KeywordExtractor}
 import org.apache.spark.streaming.StreamingContext
 import org.apache.spark.streaming.dstream.DStream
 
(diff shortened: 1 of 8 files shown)


kevinhartman:
The keyword extractor will only return exact matches. Can you explain how this is different?

There's also a requirement for conjunctive blacklist term filtering, where we filter a post only if it contains a tupleN of blacklisted terms.

I was imagining that this would be implemented by retrieving the set of blacklisted terms present using the keyword extractor, and then determining which conjunctive filter tuples were satisfied by it.

c-w:
The keyword extractor is case insensitive.

kevinhartman:
Why should the blacklist be case sensitive, but not the whitelist?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
```diff
@@ -70,6 +72,11 @@ object Pipeline {
         analysis.keywords.nonEmpty
       }
 
+      def hasBlacklistedTerms(details: Details): Boolean = {
```
kevinhartman: Ideally, this method should be added to the `Analyzer` interface, and this implementation should exist in `AnalysisDefaults`. The type of the details parameter there should be changed to accept ExtendedDetails[T] to give each concrete `Analyzer` the chance to override it, and `blacklist` will need to be added as a parameter as well.

c-w: Is there a scenario in which we do not want to blacklist terms? Also: we have a bunch of other filters that are currently not in the Analyzer trait (`hasKeywords`, `isLanguageSupported`). Do you want to include all of those too? If so: this should be a follow-up PR.

c-w:
I'd assert that it's worse to erroneously filter out an event than to erroneously include it. Ergo, we don't want any surprises for blacklist terms. For example, I may want to blacklist "Trump" (as in "the president") but not "trump" (as in "to best").

Additionally, unless I'm missing something, re-using the keyword extractor will
(1) couple us to that implementation so we won't be able to make changes to the keyword code unless they also work for the blacklist code
(2) make the code somewhat less time efficient because we have to re-tokenize the sentence once per blacklist
(3) make the code less memory efficient as we'll store the blacklist words twice
(4) make the code a little bit more complicated to understand

At the end of the day, this implementation also works, but I'm not convinced it's superior:

```scala
class Blacklist(blacklist: Seq[Set[String]]) {
  private lazy val extractors = blacklist.map(new KeywordExtractor(_))

  def matches(text: String): Boolean = {
    extractors.zip(blacklist).exists(kv => kv._1.extractKeywords(text).size == kv._2.size)
  }
}
```

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
c-w: Done in 94bf88b. Not convinced that this will or shold be ever over-written though.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/AnalysisDefaults.scala
```diff
@@ -19,6 +19,7 @@ private[analyzer] object AnalysisDefaults {
     with EnableLocation[T]
     with EnableEntity[T]
     with EnableLanguage[T]
+    with FilterBlacklist[T]
```
kevinhartman: What about `EnableBlacklist` or something that starts with `Enable` for consistency?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/topic/Blacklist.scala
```diff
@@ -0,0 +1,14 @@
+package com.microsoft.partnercatalyst.fortis.spark.transforms.topic
+
+import com.microsoft.partnercatalyst.fortis.spark.transforms.nlp.Tokenizer
+
+class Blacklist(blacklist: Seq[Set[String]]) {
```
kevinhartman, line 5: I like this since we can choose to implement this differently later on :)

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
```diff
@@ -6,16 +6,16 @@ import com.microsoft.partnercatalyst.fortis.spark.ProjectFortis.Settings
 import scala.reflect.runtime.universe.TypeTag
 import com.microsoft.partnercatalyst.fortis.spark.analyzer.{Analyzer, ExtendedFortisEvent}
 import com.microsoft.partnercatalyst.fortis.spark.dba.ConfigurationManager
-import com.microsoft.partnercatalyst.fortis.spark.dto.{Analysis, FortisEvent}
+import com.microsoft.partnercatalyst.fortis.spark.dto.{Analysis, Details, FortisEvent}
```
kevinhartman: Is this one still needed after your latest commits?

kevinhartman (approved):
LGTM

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
kevinhartman: Imagine if a pipeline had yet another field that they wanted to blacklist terms from.

I think `hasKeywords` and `isLanguageSupported` belong where they are in `Pipeline.scala`. We extract the keywords and language from the `Analyzer` since we need those results for our analysis. We don't care what the blacklisted terms were, so we don't need them as an output from the `Analyzer`.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/AnalysisDefaults.scala
c-w: 9d38b87

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
c-w: d5b30ba

PR #39, opened 2017-06-27
Deploy fat jar to Azure via Travis on tags

### LINKED ISSUE #31: Submit fortis package to maven
It looks like the artifacts for fortis-spark (and possibly others) aren't making it into maven. Definition of done: all artifacts required for one-click deploy should be published to maven.

c-w:
After merging this pull request, whenever we push a tag to Github, Travis will as part of the CI automatically build a fat jar that contains project-fortis-spark and all of its dependencies and then upload that jar to a public Azure Blob storage from where our cluster launching script can pick up the jar and pass it to spark-submit.

Sample run:
- [tag](https://github.com/CatalystCode/project-fortis-spark/tree/travis-publish-demo)
- [build](https://travis-ci.org/CatalystCode/project-fortis-spark/builds/247745513)
- [jar](https://fortiscentral.blob.core.windows.net/jars/fortis-travis-publish-demo)

This process takes about 5 minutes to run, so we could even consider running it on every Travis run (not just on tags) and name the created jars by their commit hash. This is a trivial one-line change to the `.travis/publish.sh` script so just let me know if you'd like that option.

Another thing that we can implement is to upload a second copy of the jar to a name like `fortis-latest.jar` so that the deployment script can always pick up the latest jar without having to be modified. Thoughts?

Resolves #31

### DIFF
--- a/.travis/publish.sh
+++ b/.travis/publish.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+readonly tag="${TRAVIS_TAG}"
+readonly blobaccount="${DEPLOY_BLOB_ACCOUNT_NAME}"
+readonly blobkey="${DEPLOY_BLOB_ACCOUNT_KEY}"
+readonly blobcontainer="${DEPLOY_BLOB_CONTAINER}"
+readonly blobname="fortis-${tag}.jar"
+
+log() {
+  echo "$@" >&2
+}
+
+check_preconditions() {
+  if [ -z "${tag}" ]; then
+    log "Build is not a tag, skipping publish"
+    exit 0
+  fi
+  if [ -z "${blobaccount}" ] || [ -z "${blobkey}" ] || [ -z "${blobcontainer}" ]; then
+    log "Azure blob connection is not set, unable to publish builds"
+    exit 1
+  fi
+}
+
+install_azure_cli() {
+  curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash -
+  sudo apt-get update
+  sudo apt-get install -y -qq nodejs
+  sudo npm install -g npm
+  sudo npm install -g azure-cli
+}
+
+create_fat_jar() {
+  sbt ++${TRAVIS_SCALA_VERSION} assembly
+}
+
+publish_fat_jar() {
+  local fatjar="$(find target -name 'project-fortis-spark-assembly-*.jar' -print -quit)"
+  if [ -z "${fatjar}" ] || [ ! -f "${fatjar}" ]; then
+    log "Unable to locate fat jar"
+    exit 1
+  fi
+
+  AZURE_NON_INTERACTIVE_MODE=1 \
+  azure storage blob upload \
+  --quiet \
+  --account-name "${blobaccount}" \
+  --account-key "${blobkey}" \
+  --file "${fatjar}" \
+  --container "${blobcontainer}" \
+  --blob "${blobname}"
+}
+
+check_preconditions
+install_azure_cli
+create_fat_jar
+publish_fat_jar
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 dist:
   trusty
 
+sudo:
+  required
+
 language:
   scala
 
@@ -12,3 +15,6 @@ scala:
 
 script:
   - .travis/ci.sh
+
+after_success:
+  - .travis/publish.sh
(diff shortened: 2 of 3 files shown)


.travis/publish.sh
```diff
@@ -0,0 +1,58 @@
...
+readonly blobname="fortis-${tag}.jar"
+
+log() {
+  echo "$@" >&2
+}
+
+check_preconditions() {
+  if [ -z "${tag}" ]; then
+    log "Build is not a tag, skipping publish"
+    exit 0
```
jcjimenez, line 18: Excellent!

.travis/publish.sh
```diff
@@ -0,0 +1,58 @@
...
+  fi
+  if [ -z "${blobaccount}" ] || [ -z "${blobkey}" ] || [ -z "${blobcontainer}" ]; then
+    log "Azure blob connection is not set, unable to publish builds"
+    exit 1
```
jcjimenez, line 22: Since this is runs after success, these non-zero exit codes won't block the PR, right?

.travis/publish.sh
```diff
@@ -0,0 +1,58 @@
...
+  fi
+
+  AZURE_NON_INTERACTIVE_MODE=1 \
+  azure storage blob upload \
+  --quiet \
+  --account-name "${blobaccount}" \
+  --account-key "${blobkey}" \
+  --file "${fatjar}" \
+  --container "${blobcontainer}" \
+  --blob "${blobname}"
```
jcjimenez, line 52: Any chance we could upload to blob storage via `openssl` + `curl` so we don't have to install the cli and npm?

jcjimenez (approved):
LGTM with questions.

.travis/publish.sh
c-w: Correct.

.travis/publish.sh
c-w: Unfortunately it's non-trivial to talk to Azure via curl (see this [StackOverflow post](https://stackoverflow.com/q/20103258/3817588)) so I don't think it's worth it.

.travis/publish.sh
c-w: What's your opinion on creating a fat jar for every commit instead of just tags?

jcjimenez: I want to say I prefer tags, but if we do it automatically, we may have to navigate through a bunch of junk tags in order to promote one for uploading. By the smae token, we may also end up uploading a lot of unnecessary jars. Something else we could do is to have a release branch and we merge to it when we want to upload.

c-w: Alright, let's stick with the tags then :) Thanks for your input.

.travis/publish.sh
```diff
@@ -0,0 +1,58 @@
...
+    log "Azure blob connection is not set, unable to publish builds"
+    exit 1
+  fi
+}
+
+install_azure_cli() {
+  curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash -
+  sudo apt-get update
+  sudo apt-get install -y -qq nodejs
+  sudo npm install -g npm
```
erikschlegel, line 30: Couldnt we minimize the build time by installing the `azure-cli` apt-get package as opposed to installing node? i.e. `sudo apt-get -y update && sudo apt-get install -y azure-cli`

.travis/publish.sh
erikschlegel: Not sure if you considered this, but shouldnt we  create a release in GH https://developer.github.com/v3/repos/releases/#create-a-release?

erikschlegel (commented):
This looks great @c-w . Left a couple comments below.

Thanks!

.travis/publish.sh
c-w: As far as I know, all that the Github releases button does is to create a tag. This is why the test tag that I created for this PR shows up as a release:

image

As such, we can do either (create the release on Github or a tag on the git shell) and Travis will pick it up.

.travis/publish.sh
c-w: I tried a number of ways to install the azure-cli (e.g. via apt-get or the [v2](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli#linux)) and this is the only way that I got to work reliably. The apt-get package installs a really old version of node and npm which breaks some functionality downstream.

.travis/publish.sh
c-w: Just confirmed that this works. Here is the [build](https://travis-ci.org/CatalystCode/project-fortis-spark/builds/247994555) for a release that I kicked off via the Github UI.

PR #45, opened 2017-07-05
Allow environment to override http/https for apis

c-w:
As per standup today: environment variables should be able to override whether we talk to services via http or https.


kevinhartman (approved):
LGTM

PR #47, opened 2017-07-06
Add missing Serializable on InstagramAnalyzer and Blacklist

kevinhartman:
No functional changes.

After adding these interfaces, checkpointing/recovery of our DStream graph appears to be working.


c-w (approved):
LGTM.

jcjimenez:
Post-merge question: do we need to add either of these classes to `KryoRegistrator`?

PR #53, opened 2017-07-14
Introduce sources package

c-w:
No functional change.

This change makes our project's nomenclature consistent with most stream processing projects: we have sources that generate data, transforms for processing and finally sinks for storage. Using consistent wording will make it easier for new folks to onboard and understand the project.


Smarker (approved):
LGTM

PR #55, opened 2017-07-17
Management of transform context via Service Bus

kevinhartman:
`TransformManager` listens on an Azure Service Bus Queue and asynchronously updates the set of transformation tools and state (i.e. watchlist, blacklist) used by Spark threads throughout event analysis. It acts as the gateway to this context (see `getOrUpdateContext`), ensuring the latest available context is used for each batch. 

Data fetches (e.g. from Cassandra) happen asynchronously from the perspective of Spark threads, so batch creation is not delayed while an update notification is processed (essentially, the current context is returned until it is replaced). Once data has been fetched (encapsulated in `Delta`), any required Spark broadcasts occur from Spark client threads (doing this from a worker thread would have been unexpected (Spark threads carry `threadlocal` data)).

**Note**
In adding the Service Bus client library, it was necessary to add a shadowing rule to our `sbt-assembly` configuration in order to deal with a package naming conflict between the eventhub client library (packaged into `spark-streaming-eventhubs`) and the newly added service bus client. The EH library maintainers are aware of this issue and have a fix ready, but this has not yet been released. 

The need for shadowing imposes the caveat that **only** the production artifact (fat JAR) produced by `sbt-assembly` will function properly (the target created through a regular build will hit runtime classloader errors). In the meantime (while we await EH's fix) it's still possible to debug the project locally from IDEA, but there's an extra step, which I'll document in the README.

### DIFF
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
@@ -31,22 +31,18 @@ object Pipeline {
 
     val entityModelsProvider = new ZipModelsProvider(language => s"${Settings.blobUrlBase}/opener/opener-$language.zip", Settings.modelsDir)
     val sentimentModelsProvider = new ZipModelsProvider(language => s"${Settings.blobUrlBase}/sentiment/sentiment-$language.zip", Settings.modelsDir)
-    val featureServiceClient = new FeatureServiceClient(Settings.featureServiceUrlBase)
 
     sourceStream.map(_.transform(rdd => {
       // Note: this block executes on the driver, whereas the operations applied to 'rdd' (i.e. rdd.map(_))
       // will execute on workers.
 
       // Get the shared transform context, updating it only if needed.
-      val transformContext = transformManager.getOrUpdateContext(rdd.sparkContext, configurationManager, featureServiceClient)
+      val transformContext = transformManager.getOrUpdateContext(rdd.sparkContext)
 
       // Copy TransformContext fields locally to avoid serializing everything to each task. In this way, each task's
       // serialization will only include the fields that it accesses (Spark's closure cleaner will remove the others)
       val geofence = transformContext.siteSettings.geofence
       val supportedLanguages = transformContext.siteSettings.languages
-
-      val locationsExtractorFactory = transformContext.locationsExtractorFactory
-      val locationFetcher = locationsExtractorFactory.fetch _
 
       val imageAnalyzer = transformContext.imageAnalyzer
       val languageDetector = transformContext.languageDetector
@@ -55,9 +51,10 @@ object Pipeline {
       // Broadcast variables
       val langToWatchlist = transformContext.langToWatchlist
       val blacklist = transformContext.blacklist
+      val locationsExtractorFactory = transformContext.locationsExtractorFactory
 
       def convertToSchema(original: T): ExtendedFortisEvent[T] = {
-        val message = analyzer.toSchema(original, locationFetcher, imageAnalyzer)
+        val message = analyzer.toSchema(original, locationsExtractorFactory.value.fetch _, imageAnalyzer)
         ExtendedFortisEvent(message, Analysis())
       }
 
@@ -107,7 +104,7 @@ object Pipeline {
 
       def addLocations(event: ExtendedFortisEvent[T]): ExtendedFortisEvent[T] = {
         val locations = analyzer.extractLocations(event.details,
-          locationsExtractorFactory.create(Some(new PlaceRecognizer(entityModelsProvider, event.analysis.language))))
+          locationsExtractorFactory.value.create(Some(new PlaceRecognizer(entityModelsProvider, event.analysis.language))))
         event.copy(analysis = event.analysis.copy(locations = locations))
       }
 
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
@@ -24,6 +24,11 @@ object Pipeline {
     val configs = configurationManager.fetchConnectorConfigs(name)
     val sourceStream = streamProvider.buildStream[T](ssc, configs)
 
+
+    val entityModelsProvider = new ZipModelsProvider(language => s"${Settings.blobUrlBase}/opener/opener-$language.zip", Settings.modelsDir)
+    val sentimentModelsProvider = new ZipModelsProvider(language => s"${Settings.blobUrlBase}/sentiment/sentiment-$language.zip", Settings.modelsDir)
+    val featureServiceClient = new FeatureServiceClient(Settings.featureServiceUrlBase)
+
     sourceStream.map(_.transform(rdd => {
 
       // TODO: populate these local variables with values from global update/broadcast wrappers,
@@ -31,10 +36,7 @@ object Pipeline {
       // broadcast changes across the Spark cluster.
 
       val geofence = Geofence(north = 49.6185146245, west = -124.9578052195, south = 46.8691952854, east = -121.0945042053)
-      val entityModelsProvider = new ZipModelsProvider(language => s"${Settings.blobUrlBase}/opener/opener-$language.zip", Settings.modelsDir)
-      val sentimentModelsProvider = new ZipModelsProvider(language => s"${Settings.blobUrlBase}/sentiment/sentiment-$language.zip", Settings.modelsDir)
 
-      val featureServiceClient = new FeatureServiceClient(Settings.featureServiceUrlBase)
       val locationsExtractorFactory = new LocationsExtractorFactory(featureServiceClient, geofence).buildLookup()
       val locationFetcher = locationsExtractorFactory.fetch _
       val blacklist = new Blacklist(Seq(Set("Trump", "Hilary")))
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
@@ -50,7 +50,7 @@ object Pipeline {
       val sentimentDetectorAuth = transformContext.sentimentDetectorAuth
 
       // Broadcast variables
-      val langToWatchlist = transformContext.langToWatchlist
+      val langToKeywordExtractor = transformContext.langToKeywordExtractor
       val blacklist = transformContext.blacklist
       val locationsExtractorFactory = transformContext.locationsExtractorFactory
 
@@ -74,9 +74,9 @@ object Pipeline {
       def addKeywords(event: ExtendedFortisEvent[T]): ExtendedFortisEvent[T] = {
         event.analysis.language match {
           case Some(lang) =>
-            langToWatchlist.value.get(lang) match {
-              case Some(watchlist) => event.copy(
-                analysis = event.analysis.copy(keywords = analyzer.extractKeywords(event.details, new KeywordExtractor(watchlist)))
+            langToKeywordExtractor.value.get(lang) match {
+              case Some(extractor) => event.copy(
+                analysis = event.analysis.copy(keywords = analyzer.extractKeywords(event.details, extractor))
               )
               case None => event
             }
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
@@ -4,15 +4,13 @@ package com.microsoft.partnercatalyst.fortis.spark
 import com.microsoft.partnercatalyst.fortis.spark.ProjectFortis.Settings
 import com.microsoft.partnercatalyst.fortis.spark.analyzer.{Analyzer, ExtendedFortisEvent}
 import com.microsoft.partnercatalyst.fortis.spark.dba.ConfigurationManager
-import com.microsoft.partnercatalyst.fortis.spark.dto.{Analysis, FortisEvent, Geofence}
+import com.microsoft.partnercatalyst.fortis.spark.dto.{Analysis, FortisEvent}
 import com.microsoft.partnercatalyst.fortis.spark.sources.streamprovider.StreamProvider
 import com.microsoft.partnercatalyst.fortis.spark.transforms.ZipModelsProvider
-import com.microsoft.partnercatalyst.fortis.spark.transforms.image.{ImageAnalysisAuth, ImageAnalyzer}
-import com.microsoft.partnercatalyst.fortis.spark.transforms.language.{LanguageDetector, LanguageDetectorAuth}
-import com.microsoft.partnercatalyst.fortis.spark.transforms.locations.{LocationsExtractorFactory, PlaceRecognizer}
+import com.microsoft.partnercatalyst.fortis.spark.transforms.locations.PlaceRecognizer
 import com.microsoft.partnercatalyst.fortis.spark.transforms.locations.client.FeatureServiceClient
 import com.microsoft.partnercatalyst.fortis.spark.transforms.people.PeopleRecognizer
-import com.microsoft.partnercatalyst.fortis.spark.transforms.sentiment.{SentimentDetector, SentimentDetectorAuth}
+import com.microsoft.partnercatalyst.fortis.spark.transforms.sentiment.SentimentDetector
 import com.microsoft.partnercatalyst.fortis.spark.transforms.topic.{Blacklist, KeywordExtractor}
 import org.apache.spark.streaming.StreamingContext
 import org.apache.spark.streaming.dstream.DStream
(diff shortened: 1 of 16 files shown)


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformContext.scala
```diff
@@ -0,0 +1,18 @@
...
+import com.microsoft.partnercatalyst.fortis.spark.transforms.image.ImageAnalyzer
+import com.microsoft.partnercatalyst.fortis.spark.transforms.language.LanguageDetector
+import com.microsoft.partnercatalyst.fortis.spark.transforms.locations.LocationsExtractorFactory
+import com.microsoft.partnercatalyst.fortis.spark.transforms.sentiment.SentimentDetectorAuth
+import org.apache.spark.broadcast.Broadcast
+
+case class TransformContext(
+  siteSettings: SiteSettings = null,
+  langToWatchlist: Broadcast[Map[String, List[String]]] = null,
+  blacklist: Broadcast[List[BlacklistedTerm]] = null,
```
c-w: I wonder if this should also be language-specific like the watchlist above so that we can avoid blacklisting words that happen to be the same in multiple languages. Probably a task for a follow-up PR.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
```diff
@@ -96,7 +100,7 @@ object ProjectFortis extends App {
     * Build connector config registry from hard-coded values for demo.
     *
     */
-  private object DummyConfigurationManager extends ConfigurationManager {
+  private object DummyConfigurationManager extends ConfigurationManager with Serializable {
```
c-w: Given that this is now Serializable, does it need an `@serialversionuid`?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
```diff
@@ -62,16 +71,24 @@ object Pipeline {
       }
 
       def addKeywords(event: ExtendedFortisEvent[T]): ExtendedFortisEvent[T] = {
-        val keywords = analyzer.extractKeywords(event.details, keywordExtractor)
-        event.copy(analysis = event.analysis.copy(keywords = keywords))
+        event.analysis.language match {
+          case Some(lang) =>
+            langToWatchlist.value.get(lang) match {
+              case Some(watchlist) => event.copy(
+                analysis = event.analysis.copy(keywords = analyzer.extractKeywords(event.details, new KeywordExtractor(watchlist)))
```
c-w: It's not ideal that we re-instantiate the KeywordExtractor here all the time as the building of its internal Trie for sure must create a lot of garbage as it allocates many strings. I wonder if we can work around it by having a lazy wrapper around the langToWatchlist that generate each keyword extractor just once.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
```diff
@@ -205,7 +209,31 @@ object ProjectFortis extends App {
...
+    override def fetchSiteSettings(): SiteSettings = SiteSettings(
+      id = null,
+      siteName = "",
+      geofence = Geofence(north = 49.6185146245, west = -124.9578052195, south = 46.8691952854, east = -121.0945042053),
+      languages = Set("en", "fr", "de"),
+      defaultZoom = 0,
+      title = "",
+      logo = "",
+      translationSvcToken = Settings.oxfordLanguageToken,
+      cogSpeechSvcToken = System.getenv("OXFORD_SPEECH_TOKEN"),
```
c-w, line 225: For consistency, this should probably move to Settings like the other tokens. Fine to do as a follow-up PR as this one is pretty big already.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
c-w: Same for TransformManager below.

jcjimenez (approved):
LGTM

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
```diff
@@ -0,0 +1,256 @@
...
+      finally {
+        writeLock.unlock()
+      }
+    }
+  }
+
+  private def startQueueClient(): Unit = {
+    queueClient = new QueueClient(
+      new ConnectionStringBuilder(
+        Properties.envOrElse("FORTIS_SERVICEBUS_NAMESPACE", ""),
```
c-w: Should probably fail early in case the required environment variables are not set.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
```diff
@@ -0,0 +1,256 @@
...
+
+  /**
+    * Builds a delta against the current transform context using the provided site settings.
+    * @param delta An existing delta to be used as a base.
+    * @return
+    */
+  private def deltaWithSettings(siteSettings: SiteSettings, delta: Delta = Delta()): Delta = {
+    // Note that parameters are call-by-name
+    def updatedField[T](isDirty: => Boolean, newVal: => T): Option[T] = {
+      // Invoke 'isDirty' only if transform context is not null
```
c-w: We can avoid this comment and make it explicit in the code which may be more readable:

```scala
if (transformContext == null) {
  Some(newVal)
} else {
  isDirty()
  None
}
```

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
```diff
@@ -0,0 +1,256 @@
...
+      ),
+      sentimentDetectorAuth = updatedField(
+        siteSettings.translationSvcToken != transformContext.siteSettings.translationSvcToken,
+        SentimentDetectorAuth(siteSettings.translationSvcToken)
+      )
+    )
+  }
+
+  /**
+    * Builds a delta against the current transform context using the provided watchlist.
```
c-w: I don't understand this comment; it doesn't seem to reflect the code. The comment mentions a "current transform context", but the method itself could be made static, so how does it reference state as implied by the comment?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
```diff
@@ -0,0 +1,256 @@
...
+
+        throw new Exception("No Spark thread acknowledged the update message within the timeout.")
+      }
+
+      CompletableFuture.completedFuture(null)
+    }
+  }
+}
+
+private object TransformManager {
```
c-w: Is this wrapper required? What value does it add on top of having the delta as a top level class?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
```diff
@@ -0,0 +1,256 @@
...
+
+  /**
+    * Holds the next set of values for the fields of the current transform context. If the value of a field here is
+    * empty, then the corresponding field of the transform context during update will be left as is.
```
c-w: I wonder if this can be made more explicit (and therewith the comment redundant_ by having a top-level constant called `NoChange` that's an alias for `None`.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
```diff
@@ -0,0 +1,256 @@
...
+          delta = deltaWithBlacklist(blacklist, delta)
+
+          // Create transformContext and broadcast
+          updateTransformContext(delta, sparkContext)
+
+          // Start async update listener
+          startQueueClient()
+        }
+      }
+      finally {
```
c-w: Nit: elsewhere puts the keyword on the same line as the closing brace.

```scala
} finally {
```

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
```diff
@@ -0,0 +1,256 @@
...
+          // Fetch data synchronously
+          val siteSettings = configManager.fetchSiteSettings()
+          val langToWatchlist = configManager.fetchWatchlist()
+          val blacklist = configManager.fetchBlacklist()
+
+          var delta = deltaWithSettings(siteSettings)
+          delta = deltaWithWatchlist(langToWatchlist, delta)
+          delta = deltaWithBlacklist(blacklist, delta)
+
+          // Create transformContext and broadcast
```
c-w: Comment can be made redundant by renaming method below to `updateTransformContextAndBroadcast`

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
```diff
@@ -0,0 +1,256 @@
...
+      writeLock.lock()
+      try {
+        if (transformContext == null) {
+
+          // Fetch data synchronously
+          val siteSettings = configManager.fetchSiteSettings()
+          val langToWatchlist = configManager.fetchWatchlist()
+          val blacklist = configManager.fetchBlacklist()
+
+          var delta = deltaWithSettings(siteSettings)
```
c-w: Why not just use this instead of creating three objects?

```scala
var delta = Delta(siteSettings = siteSettings, langToWatchlist = langToWatchlist, blacklist = blacklist)
```

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
```diff
@@ -0,0 +1,256 @@
...
+          delta = deltaWithWatchlist(langToWatchlist, delta)
+          delta = deltaWithBlacklist(blacklist, delta)
+
+          // Create transformContext and broadcast
+          updateTransformContext(delta, sparkContext)
+
+          // Start async update listener
```
c-w: Comment is redundant with code. Refer to [this article](https://ncrcoe.gitbooks.io/java-for-small-teams/style/20_prefer_readable_code_to_comments.html) for a good philosophy on commenting.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
```diff
@@ -0,0 +1,256 @@
...
+        Properties.envOrElse("FORTIS_SERVICEBUS_CONFIG_QUEUE", ""),
+        Properties.envOrElse("FORTIS_SERVICEBUS_POLICY_NAME", ""),
+        Properties.envOrElse("FORTIS_SERVICEBUS_POLICY_KEY", "")
+      ), ReceiveMode.PeekLock)
+
+    queueClient.registerMessageHandler(
+      new MessageHandler(),
+      new MessageHandlerOptions(
+        1, // Max concurrent calls
+        true,
```
c-w: Would it be possible to use named arguments here (and for the duration below) so we don't have to guess what this flag refers to? Similarly, named arguments will also make the comment for the integer above redundant.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformContext.scala
```diff
@@ -0,0 +1,18 @@
...
+  locationsExtractorFactory: Broadcast[LocationsExtractorFactory] = null,
+  imageAnalyzer: ImageAnalyzer = null,
```
c-w: Would it be worth a divider comment to denote the following fields as "cheap to create so not broadcastet" to make it obvious why only some of the fields in this object are broadcasts?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
```diff
@@ -0,0 +1,256 @@
...
+import com.microsoft.partnercatalyst.fortis.spark.transforms.image.{ImageAnalysisAuth, ImageAnalyzer}
+import com.microsoft.partnercatalyst.fortis.spark.transforms.language.{LanguageDetector, LanguageDetectorAuth}
+import com.microsoft.partnercatalyst.fortis.spark.transforms.locations.LocationsExtractorFactory
+import com.microsoft.partnercatalyst.fortis.spark.transforms.locations.client.FeatureServiceClient
+import com.microsoft.partnercatalyst.fortis.spark.transforms.sentiment.SentimentDetectorAuth
+import org.apache.spark.SparkContext
+
+import scala.util.Properties
+
+class TransformManager(configManager: ConfigurationManager, featureServiceClient: FeatureServiceClient) extends Serializable with Loggable {
```
c-w: The "Manager" suffix is one of those [weasel words](https://blog.codinghorror.com/i-shall-call-it-somethingmanager/)... I wonder if we can come up with a more descriptive name?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
```diff
@@ -0,0 +1,256 @@
...
+  private val shouldUpdate: SynchronousQueue[Delta] = new SynchronousQueue[Delta]()
+  private val writeLock: ReentrantLock = new ReentrantLock(true)
+
+  // Do not serialize these values. The transformContext would otherwise contain stale Broadcast
+  // instances, so we rebuild it on recovery from checkpoint.
+  @volatile @transient private var transformContext: TransformContext = null
+  @volatile @transient private var queueClient: QueueClient = null
+
+  def getOrUpdateContext(sparkContext: SparkContext): TransformContext = {
```
c-w: The manager always needs a context to function, so would it be possible to pass it as a field via the constructor? That would also let us get rid of the initializer function in favor of a lazy initialized field which is safer as the developer doesn't have to make sure to always call the initializer function before accessing the field.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
```diff
@@ -0,0 +1,256 @@
...
+    ensureInitialized(sparkContext)
+
+    writeLock.lock()
+    try {
+      // Grab delta from hand-off only if it's available *now*
+      val delta = Option(shouldUpdate.poll(0, TimeUnit.SECONDS))
+
+      // Update transform context with delta
```
c-w: This comment would be made unnecessary if we avoided the option/foreach below in favor of something more explicit like this:

```scala
if (delta != null) {
  updateTransformContext(delta, sparkContext)
}
```

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
c-w: I wonder if this class wants to be split into two for simplicity: one class for receiving deltas (in this current implementation via ServiceBus) and one for updating the broadcast with the delta.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
```diff
@@ -0,0 +1,256 @@
...
+  // instances, so we rebuild it on recovery from checkpoint.
+  @volatile @transient private var transformContext: TransformContext = null
+  @volatile @transient private var queueClient: QueueClient = null
+
+  def getOrUpdateContext(sparkContext: SparkContext): TransformContext = {
+    ensureInitialized(sparkContext)
+
+    writeLock.lock()
+    try {
+      // Grab delta from hand-off only if it's available *now*
```
c-w: This comment tells me what the code below is doing, but not why. It would be very useful to know why it's important that the hand-off is available right now.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
```diff
@@ -0,0 +1,256 @@
...
+    transformContext
+  }
+
+  private def ensureInitialized(sparkContext: SparkContext): Unit = {
+    if (transformContext == null) {
+      // Blocking init of transform context and queue client used for non-blocking updates.
+      // Initialization has not yet occurred since construction/deserialization.
+      writeLock.lock()
+      try {
+        if (transformContext == null) {
```
c-w: This is essentially implementing double check locking which is easily broken in many memory models (sample [article for .NET](http://csharpindepth.com/Articles/General/Singleton.aspx#dcl)), similar resources exist for [the JVM](https://stackoverflow.com/a/25262497/3817588). It would be simpler and safer to use a `lazy` field instead.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
```diff
@@ -0,0 +1,256 @@
...
+            deltaWithWatchlist(langToWatchlist)
+          case "blacklist" =>
+            val blacklist = configManager.fetchBlacklist()
+            deltaWithBlacklist(blacklist)
+          case unknown =>
+            logError(s"Service Bus client received unexpected update request. Ignoring.: $unknown")
+            Delta()
+        }
+        case None =>
+          logError(s"Service Bus client received unexpected message. Ignoring.: ${message.toString}")
```
c-w: Isn't the toString implied by it being inside of a string interpolation?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  /**
+    * Builds a delta against the current transform context using the provided watchlist.
+    * @param delta An existing delta to be used as a base.
+    * @return
+    */
+  private def deltaWithWatchlist(langToWatchlist: Map[String, List[String]], delta: Delta = Delta()) = {
+    delta.copy(
+      langToWatchlist = Some(langToWatchlist)
```
c-w: This only lets us replace the entire watchlist in one go. Via the dashboard we'll often be adding single terms, so I wonder if this should instead update the watchlist terms instead of replacing them, so that we don't have to fetch the entire watchlist from the DB. More generally, whenever a message arrives, with the current approach we'll have to fetch the new values from the DB -- why not just put the delta into the queue message directly? Sure, message sizes are limited, but will we really hit them if all we do is update single words? Why refetch data that we already have available :).

c-w (dismissed):
Good start. I think that the TransformManager may have two or three smaller classes inside of it though.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
kevinhartman: Unfortunately we cannot do this since the `SparkContext` is not serializable. The `TransformManager` class will be serialized along with the DStream graph during checkpointing.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
kevinhartman: Double check locking should be fine as per recent JMM revision implementations. The `volatile` declaration of `transformContext` should ensure that the construction of the `TransformContext` instance "happens-before" the reference is updated.

This is also used in the [official Spark Streaming example](https://github.com/apache/spark/blob/master/examples/src/main/scala/org/apache/spark/examples/streaming/RecoverableNetworkWordCount.scala#L40).

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
kevinhartman: My thought is that we should ditch the whole `Settings` thing and just hit env vars at time of usage. We can add early fails somewhere in the main. I'll add constants for these env var names for now.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
kevinhartman: Unfortunately, no. Decorating these with parameter names fails to compile. The comment is my workaround ;)

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
kevinhartman: That's not semantically equivalent. The idea is that `isDirty` is not evaluated as part of the conditional unless the transform context is not null. I'm taking advantage of the early out of ||. Also note that `isDirty` is a named expression, and not invokable (). If this is too cryptic, I can rework or improve the comment.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
kevinhartman: It was initially here for symmetry with `deltaWithSettings` to communicate that the implementation may reference the context in the future. In practice, I think it'd be nonsensical, but at the time I didn't bother exploring the idea. I'll remove the comment.

kevinhartman: Well actually, if we end up updating per-word, this function can copy the existing `langToWatchlist`, adding/removing a word.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
kevinhartman: That'd be a nice improvement that we could easily add on top of the current implementation in a future changeset. For now, I've favored the atomicity of fetching the entire set at a time given the current state of the Service Bus client library.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
kevinhartman: Yes. For better or worse, I'd added it explicitly to allude to `message` not being a string.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
kevinhartman: In Java, this amounts to adding `Delta` as a static inner class.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
kevinhartman: As much as I like the idea of building tiny DSLs (>:3), I'm not convinced that would improve readability. I'll try it (or something similar), but I'm not expecting to introduce a change.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
kevinhartman: In this case, I don't think `Manager` is too bad :p 

How about `TransformContextProvider`?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/TransformManager.scala
kevinhartman: Generally, I agree. In this case, I chose to do this for consistency. If this were the only major step within this block without a comment, the reader may infer that it's less important or a less certain statement. Fixing the update method's name as you've suggested (`updateTransformContextAndBroadcast`) should justify this comment's removal as well.

PR #57, opened 2017-07-20
Migrating Period and PeriodType from playground.

jcjimenez:
Starting to migrate some things from my playground. These classes would be required for creating an analysis request. These requests would then live in something like a Set[Analysis] request to be processed when the period is mature enough for processing.


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/Area.scala
```diff
@@ -0,0 +1,44 @@
+package com.microsoft.partnercatalyst.fortis.spark.analyzer.timeseries
+
+trait AreaFinder {
+  def findArea(tilez: Int, tilex: Int, tiley: Int): Option[Area]
```
jcjimenez: @erikschlegel this trait may already exist somewhere in your code. It represents a black box which takes tiles coordinates and gives you a general are for which it belongs. A simple implementation would be to have the coordinates `mod` to a tile size and a more complex one could use something like clustering to determine more appropriate area given existing activity.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/Area.scala
```diff
@@ -0,0 +1,44 @@
...
+    Area(
+      tilez,
+      Tuple2(Math.min(this.rangex._1, that.rangex._1), Math.max(this.rangex._2, that.rangex._2)),
+      Tuple2(Math.min(this.rangey._1, that.rangey._1), Math.max(this.rangey._2, that.rangey._2))
+    )
+  }
+
+  def intersectsWith(other: Area): Boolean = {
+    if (this.tilez != other.tilez) {
+      return false
```
Smarker: Rather than using `return` for some of these, can you instead use `if, else if, else` so you don't need the `return`?

Smarker (approved):
LGTM

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/Area.scala
jcjimenez: I think you are right in that early returns aren't very Scala-like (or functional in style). But, for some reason, early returns feel to me like the more readable approach here, e.g. "different tilez values? stop now" whereas if/else chains have a slightly-higher cognitive cost because one has to look at the whole expression.

PR #59, opened 2017-07-21
Analyzer refactor so fortisEvents are inline with cassandra schema

erikschlegel:
This PR encompasses a variety of refactoring changes to ensure the naming conventions are consistent with Cassandra events table. Also, had to change the event time data source for all analyzers so it's based off the upstream event time as opposed to the current system time.


jcjimenez (approved):
LGTM

PR #61, opened 2017-07-24
Unplugged trend (not anomaly) detection. Need to connect to CassandraSink.

jcjimenez:
This is roughly how the trend computation would be connected (assuming the topic stream is throttled in a way that is compatible with this):

```
    conjunctionTopicStream.transform(rdd=>{
      val topics: Set[(String, String, String)] = rdd.collect().toSet
      val tilesDAO = new ComputedTilesDAO()
      (8 to 16).par.flatMap(tilez=>{
        PeriodType.all.par.flatMap(periodType=>{
            ComputedTilesToComputedTrends(tilesDAO.latestTiles(
              rdd.sparkContext,
              topics,
              periodType,
              new Date().getTime,
              Seq(tilez)
            ))
        })
      })
      trendsPerLevel.reduce((rdd1, rdd2)=>rdd1.union(rdd2))
    }).foreachRDD(trendRDD=>{
      trendRDD.saveToCassandra("fortis", "computedtrends")
    })
```

Once this data is in, the UI would leverage it (via our GraphQL endpoints) with a query similar to this one:

```
select * from computedtrends where pipelinekey = 'all' and periodtype = 'day' and period = '2017-06-20' and tilez = 1 and (tilex, tiley) >= (0, 0) and (tilex, tiley) < (10, 10);
```


build.sbt
```diff
@@ -14,8 +14,11 @@ scalacOptions ++= Seq(
 
 val sparkVersion = "2.1.0"
 
+parallelExecution in Test := false
+
```
jcjimenez, line 18: I had to disable parallel execution of tests (for now) since the two tests relying on SparkContext/StreamingContext were stepping over each other's toes. A better approach may be to move these kinds of tests into a suite that's configured to run its tests sequentially.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/Period.scala
```diff
@@ -35,7 +35,8 @@ sealed class PeriodType(val sizeInMilliseconds: Long, format: String, truncateFi
     if (timeZone.isDefined) {
       sdf.setTimeZone(timeZone.get)
```
erikschlegel: Add UTC default timezone

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/Period.scala
```diff
@@ -3,28 +3,28 @@ package com.microsoft.partnercatalyst.fortis.spark.analyzer.timeseries
...
+  val Hour = PeriodType("hour", 60L * Minute.sizeInMilliseconds, "yyyy-MM-dd HH", Set(Calendar.MINUTE))
 
-  case object Day extends PeriodType(24L * Hour.sizeInMilliseconds, "yyyy-MM-dd", Set(Calendar.MINUTE, Calendar.HOUR_OF_DAY))
+  val Day = PeriodType("day", 24L * Hour.sizeInMilliseconds, "yyyy-MM-dd", Set(Calendar.MINUTE, Calendar.HOUR_OF_DAY))
 
-  case object Month extends PeriodType(30L * Day.sizeInMilliseconds, "yyyy-MM", Set(Calendar.MINUTE, Calendar.HOUR_OF_DAY, Calendar.DAY_OF_MONTH))
+  val Week = PeriodType("week", 7L * Day.sizeInMilliseconds, "yyyy-w", Set(Calendar.MINUTE, Calendar.HOUR_OF_DAY, Calendar.WEEK_OF_YEAR))
 
-  case object Year extends PeriodType(12L * Month.sizeInMilliseconds, "yyyy", Set(Calendar.MINUTE, Calendar.HOUR_OF_DAY, Calendar.DAY_OF_MONTH, Calendar.MONTH))
+  val Month = PeriodType("month", 30L * Day.sizeInMilliseconds, "yyyy-MM", Set(Calendar.MINUTE, Calendar.HOUR_OF_DAY, Calendar.DAY_OF_MONTH))
```
erikschlegel: Compute number of milliseconds based on month of year.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTilesDAO.scala
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.SparkSession
+
+trait ComputedTilesDAOConfig extends Serializable {
+  def retrospectivePeriodCountPerType(): Map[String, Int]
+}
+
+case class DefaultComputedTilesDAOConfig() extends ComputedTilesDAOConfig {
+  override def retrospectivePeriodCountPerType(): Map[String, Int] = Map(
+    (PeriodType.Minute.periodTypeName, 4*60), /* 4 hours */
```
c-w: I don't understand what this config does. Why would we map "minute" to 4 hours?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTilesDAO.scala
```diff
@@ -0,0 +1,67 @@
...
+      "tiley",
+      "externalsourceid",
+      "mentioncount",
+      "avgsentiment",
+      "heatmap",
+      "placeids",
+      "insertion_time",
+      "conjunctiontopics"
+    ).where(
+      "conjunctiontopics in ? and periodtype = ? and tilez in ? and period in ?",
```
c-w: We should keep an eye on these `IN` clauses as Cassandra is not very good at pulling data from multiple partitions. Rule of thumb usually is that it's faster to make multiple queries in the client in parallel.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTilesToComputedTrends.scala
```diff
@@ -0,0 +1,45 @@
...
+    val groupedTiles = computedTiles.groupBy(tileGroup)
+    groupedTiles.flatMap[ComputedTrend]((group) => {
+      val tiles = group._2.toSeq
+
+      val regressionResult = linearRegression(tiles)
+
+      val score = regressionResult.getSlope
+      if (score.isNaN) Seq[ComputedTrend]()
+      else {
+        val latestTile = tiles.sortBy(t => -t.periodstartdate).head
```
c-w: Would this be clearer? `tiles.maxBy(_.periodstartdate)` (Minor added benefit: O(n) instead of O(n log n).)

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTilesToComputedTrends.scala
```diff
@@ -0,0 +1,45 @@
...
+        val filteredTiles = tiles.filter(t => t.period == latestTile.period)
```
c-w: Is there a more descriptive name than `filteredTiles`? We already know they're filtered from the filter call on the right hand side. E.g. `tilesInLatestPeriod`?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTilesToComputedTrends.scala
```diff
@@ -0,0 +1,45 @@
...
+        val trends = filteredTiles.map(new ComputedTrend(_, score))
+        trends
+      }
+    })
+  }
+
+  def tileGroup(computedTile: ComputedTile): String = {
+    (
+      computedTile.pipelinekey,
```
c-w, line 29: Might as well us an s-string to make more explicit that we're making a compound key here. `s"${computedTile.piplinekey}_${computedTile.conjunctiontopics}_${computedTile.tilez}"`. Added bonus: no need for explicit toString calls. In this simplified way, we could also just inline the grouping condition to make the groupby call more immediately understandable without having to jump to this definition.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTilesToComputedTrends.scala
```diff
@@ -0,0 +1,45 @@
...
+      computedTile.conjunctiontopics.toString,
+      computedTile.tilez.toString
+    ).toString()
+  }
+
+  def linearRegression(computedTiles: Seq[ComputedTile]): SimpleRegression = {
+    computedTiles.par.aggregate(new SimpleRegression(true))((regression, tile)=>{
```
c-w, line 36: Could we use a [named argument](http://docs.scala-lang.org/tutorials/tour/named-arguments.html) so that it's more obvious what the `true` does without having to jump to the regression docs? In this case, `SimpleRegression(includeIntercept=true)`

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/Period.scala
```diff
@@ -35,7 +35,8 @@ sealed class PeriodType(val sizeInMilliseconds: Long, format: String, truncateFi
     if (timeZone.isDefined) {
       sdf.setTimeZone(timeZone.get)
     }
-    sdf.format(date)
+    val formattedDate = sdf.format(date)
+    this.periodTypeName + "-" + formattedDate
```
c-w: Can s-strings make this more readable? `s"${this.periodTypeName}-$formattedDate"`

src/test/scala/com/microsoft/partnercatalyst/fortis/spark/SparkContextCoordinator.scala
```diff
@@ -0,0 +1,31 @@
+package com.microsoft.partnercatalyst.fortis.spark
+
+import org.apache.spark.streaming.{Seconds, StreamingContext}
+import org.apache.spark.{SparkConf, SparkContext}
+
+object SparkContextCoordinator {
```
c-w: There is a project called [spark-testing](https://github.com/holdenk/spark-testing-base) which has a bunch of smart base classes for unit testing Spark apps. I looked into this in the past but unfortunately they don't have streaming support so I dropped the investigation. Given that we now have a few use-cases for Spark Streaming tests, I wonder if we should look into contributing a streaming test base class to their project. Worth discussing during standup and I'm happy to take the task if we decide to go down this route.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTilesToComputedTrends.scala
```diff
@@ -0,0 +1,45 @@
...
+      val tiles = group._2.toSeq
+
+      val regressionResult = linearRegression(tiles)
+
+      val score = regressionResult.getSlope
+      if (score.isNaN) Seq[ComputedTrend]()
+      else {
+        val latestTile = tiles.sortBy(t => -t.periodstartdate).head
+        val filteredTiles = tiles.filter(t => t.period == latestTile.period)
+        val trends = filteredTiles.map(new ComputedTrend(_, score))
```
c-w: I'm not sure what we're doing here. From my understanding, the idea for the anomaly detection was to fit a linear model to the mention counts and then flag any data point that has a high error with regards to the model as an anomaly, i.e. trend. The code here instead fits the regression, discards its intercept and tags every data point with the slope of the overall regression. What will this value be used for?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTilesToComputedTrends.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.spark.rdd.RDD
+
+object ComputedTilesToComputedTrends {
+
+  def apply(computedTiles: RDD[ComputedTile]): RDD[ComputedTrend] = {
+    val groupedTiles = computedTiles.groupBy(tileGroup)
+    groupedTiles.flatMap[ComputedTrend]((group) => {
+      val tiles = group._2.toSeq
+
+      val regressionResult = linearRegression(tiles)
```
c-w, line 14: Just as a general point: using linear regression for anomaly detection is very naive, to the point of not being useful in many real-world situations. From my understanding, there are three types of timeseries anomalies that we normally care about: spikes, step changes and shifts. With a standard linear model, we can at best hope to detect spikes.

A still simple but slightly less naive model would incorporate at least a trend term (e.g. by rolling median for the last N timeframes) and a seasonality term (e.g. via fast-fourier on the last N timeframes). There is a good [article by AirBnB](https://medium.com/airbnb-engineering/anomaly-detection-for-airbnb-s-payment-platform-e3b0ec513199) on a simple model like this.

c-w (dismissed):
Would love it if the code could include some more context on the implementation goals and assumptions of the trend detection so that other folks can dive into this in the future.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTilesDAO.scala
jcjimenez: Good question. The objective of this is to answer questions like "how far back should I go if I'm making a by-the-minute analysis" and in this case, this map would give an answer of "240 minutes" (4 hours).

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTilesDAO.scala
jcjimenez: Good observation. In general, as I put in this PR's comment, we will be performing those queries in parallel (with the exception of `period`). Maybe the other `in` clauses should be converted to `=` since they are most-likely `YAGNI`.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTilesToComputedTrends.scala
jcjimenez: You are correct. This isn't detecting anomalies, but general trends. Stay tuned for the anomalies code (in a different PR).

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTilesToComputedTrends.scala
jcjimenez: Again, this is just plotting trends, not anomalies.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTilesToComputedTrends.scala
jcjimenez: You make a reasonable point, but I wanted to factor this out to have a test to illustrate what the groups look like:

```
  it should "format groups" in {
    val group = ComputedTilesToComputedTrends.tileGroup(
      ComputedTile(2017, 0, PeriodType.Year.periodTypeName, "2017", "twitter", 10, 1, 2, "HamillHimself", 150, 0, "", Seq(), 0, Tuple3(Some("europe"), None, None))
    )
    assert(group == "(twitter,(Some(europe),None,None),10)")
  }
```

Happy to remove this if you feel  strongly about it.

src/test/scala/com/microsoft/partnercatalyst/fortis/spark/SparkContextCoordinator.scala
jcjimenez: I meant to omit this from my PR. Apologies for the confusion.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/Period.scala
jcjimenez: In some cases, I would agree wrt the readability of the formatted string, but I think this case seems more readable to me with simple concatenation (and saves the minor performance hit of the string interpolation).

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTilesToComputedTrends.scala
jcjimenez: Not a bad idea. However, since this class is not native Scala, it won't let me include the named arg.

erikschlegel (approved):
LGTM

PR #63, opened 2017-07-31
Preliminary changes for kmeans-based anomaly detection.

jcjimenez:
For review purposes at the moment - DO NOT MERGE.

For the purposes of this explanation, consider the data as having 2 features, topic count and time period. However, in practice, two more features are taken into account: the pipeline key hash value (e.g. the hash value for `"twitter"`) and the external source id hash value (e.g. `"HamillHimself"`)

This implementation relies on first transforming raw time series data like the following:

topic_count_to_date

Into a spectrum/histogram whose data is separated into Fortis' existing period types (`day`, `week`, `month`, `year`). The following day-of-week spectrum (x axis 1: Monday, 2: Tuesday etc.) is an example of the data passed into the k-means cluster model for fitting:

topic_count_by_day_of_week

Such a transformation is then passed through a kmeans cluster whose `k` value is proportional to the number of values in the x axis. Also, the feature values in the x axis are multiplied by a factor large enough to help spread out the clusters.

Lastly, when a new topic count (`ComputedTile`) is encountered, it is passed through each of the k-means models, and if its cost is greater than the model cost (sum of squared distances of points to their nearest center), it is considered an anomaly for that period.

Also, I just watched this video and noticed they are doing something similar to what I ended up doing (histogram-based approach with DBSCAN):

https://www.youtube.com/watch?v=wacqvGVXuxc

### DIFF
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedAnomalyKmeansInputRecord.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedAnomalyKmeansInputRecord.scala
@@ -1,0 +1,29 @@
+package com.microsoft.partnercatalyst.fortis.spark.analyzer.timeseries
+
+import com.microsoft.partnercatalyst.fortis.spark.dto.TopicCount
+import org.apache.spark.ml.linalg.{Vector, Vectors}
+
+case class ComputedAnomalyKmeansInputRecord(
+  period: String,
+  periodtype: String,
+  pipelinekey: String,
+  externalsourceid: String,
+  features: Vector
+) extends Serializable
+
+object ComputedAnomalyKmeansInputRecord {
+
+  def apply(topicCount: TopicCount,
+            parameters: ComputedAnomalyKmeansModelParameter): ComputedAnomalyKmeansInputRecord = ComputedAnomalyKmeansInputRecord(
+    topicCount.period,
+    topicCount.periodtype,
+    topicCount.pipelinekey,
+    topicCount.externalsourceid,
+    Vectors.dense(
+      parameters.periodFeature(topicCount),
+      topicCount.pipelinekey.hashCode.toDouble,
+      topicCount.externalsourceid.hashCode.toDouble,
+      topicCount.mentioncount.toDouble)
+  )
+
+}
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedAnomalyKmeansModelInputDStream.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedAnomalyKmeansModelInputDStream.scala
@@ -1,21 +1,19 @@
 package com.microsoft.partnercatalyst.fortis.spark.analyzer.timeseries
 
-import java.time.{ZoneId, ZonedDateTime}
 import java.util.Date
 
 import com.datastax.spark.connector._
 import com.microsoft.partnercatalyst.fortis.spark.dto.{MLModel, TopicCount, TrustedSource}
 import org.apache.spark.ml.clustering.KMeans
-import org.apache.spark.ml.linalg.{Vector, Vectors}
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.SparkSession
-import org.apache.spark.streaming.dstream.InputDStream
+import org.apache.spark.streaming.dstream.{DStream, InputDStream}
 import org.apache.spark.streaming.{Duration, Minutes, StreamingContext, Time}
 
 /**
-  * Given that Fortis already has a notion of Frequency (i.e. different kinds of period types), detecting signal
-  * frequencies (e.g. using an FFT or other methods) isn't necessary (could be useful, but not must-have). Leveraging
-  * these period types, they can be used as implicit k-means cluster anchors.
+  * Given that Fortis has a notion of Frequency (i.e. different kinds of period types), detecting signal frequencies
+  * (e.g. using an FFT or other methods) isn't necessary (could be useful, but not must-have). Leveraging these period
+  * types, they can be used as implicit k-means cluster anchors.
   *
   * So, in theory, the normal behavior of a given topic could be represented as a set of histograms (e.g. one such
   * histogram could have x axis values of Sunday, Monday, Tuesday, Wednesday, Thursday, Friday, and Saturday). Such
@@ -23,25 +21,25 @@ import org.apache.spark.streaming.{Duration, Minutes, StreamingContext, Time}
   *
   * Because having a model for each pipeline+user combination may be impractical (both because of the number of models
   * and because of how one would query Cassandra to build these), two dimensions are included in the feature space:
-  * pipelinekey and externalsourceid. The hashcode of these two strings is used as their feature values (instead of
+  * pipelinekey and externalsourceid. The hashCode of these two strings is used as their feature values (instead of
   * using something like org.apache.spark.ml.feature.StringIndexer) because the hashcode values gives enough
   * [potential] spacing between features (as opposed to values from StringIndexer which are represented by a sequence
   * of a step of 1).
   */
 object ComputedAnomalyKmeansModelInputDStream {
 
-  def apply(periodType: PeriodType, ssc: StreamingContext): ComputedAnomalyKmeansModelInputDStream = new ComputedAnomalyKmeansModelInputDStream(periodType, ssc)
-
-  def preferedStreamingDuration(periodType: PeriodType): Duration = {
-    periodType match {
-      case PeriodType.Minute => Minutes(15)
-      case PeriodType.Hour => Minutes(30)
-      case PeriodType.Day => Minutes(60)
-      case PeriodType.Month => Minutes(4*60)
-      case PeriodType.Year => Minutes(12*60)
-      case _ => Minutes(60)
-    }
+  def apply(ssc: StreamingContext): DStream[MLModel] = {
+    ssc.union(Seq(
+      ComputedAnomalyKmeansModelInputDStream(PeriodType.Minute, ssc),
+      ComputedAnomalyKmeansModelInputDStream(PeriodType.Hour, ssc),
+      ComputedAnomalyKmeansModelInputDStream(PeriodType.Day, ssc),
+      ComputedAnomalyKmeansModelInputDStream(PeriodType.Week, ssc),
+      ComputedAnomalyKmeansModelInputDStream(PeriodType.Month, ssc),
+      ComputedAnomalyKmeansModelInputDStream(PeriodType.Year, ssc)
+    ))
   }
+
+  def apply(periodType: PeriodType, ssc: StreamingContext): ComputedAnomalyKmeansModelInputDStream = new ComputedAnomalyKmeansModelInputDStream(periodType, ssc)
 
 }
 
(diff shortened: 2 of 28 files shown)


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedAnomalyKmeansModelInputDStream.scala
```diff
@@ -0,0 +1,149 @@
...
+  )
+
+}
+
+private object ComputedAnomalyKmeansInputRecord extends Serializable {
+
+  def getK(periodType: PeriodType,
+           uniqueSourcesCount: Int,
+           uniqueTopicsCount: Int): Int = {
+    var k = 31 /* max days in a month */
```
jcjimenez: Do not be spooked by these hard-coded, period-specific, values. I'm in the process of generalizing this for the different period types. Some period types, like `day`, will have multiple models (e.g. daily for day of the week, day of the month, and day of the year).

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/dto/TrustedSource.scala
```diff
@@ -0,0 +1,7 @@
+package com.microsoft.partnercatalyst.fortis.spark.dto
+
+case class TrustedSource(pipelinekey: String,
```
c-w: Nit: elsewhere we use this style for case classes:

```scala
case class TrustedSource(
  pipelinekey: String,
  externalsourceid: String,
  sourcetype: String,
  rank: Int,
  insertion_time: Long
) extends Serializable
```

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/dto/TopicCount.scala
```diff
@@ -0,0 +1,13 @@
+package com.microsoft.partnercatalyst.fortis.spark.dto
+
+case class TopicCount(periodtype: String,
+                      period:String,
```
c-w: Nit: space after colon.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/dto/MLModel.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream, ObjectInputStream, ObjectOutputStream}
+import java.util.Date
+
+object MLModel {
+
+  def dataFrom(modelObj: Serializable): Array[Byte] = {
+    val stream = new ByteArrayOutputStream()
+    val oos = new ObjectOutputStream(stream)
+    oos.writeObject(modelObj)
```
c-w: Apache-Commons has a nice [helper](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/SerializationUtils.html#serialize-java.io.Serializable-) for this serialization and deserialization:

```scala
import org.apache.commons.lang.SerializationUtils.serialize;
return serialize(modelObj);
```

In case you want to use this:

```scala
// https://mvnrepository.com/artifact/commons-lang/commons-lang
libraryDependencies += "commons-lang" % "commons-lang" % "2.6"
```

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedAnomalyKmeansModelInputDStream.scala
```diff
@@ -0,0 +1,149 @@
...
+    val uniqueSourcesCount = topicCounts.toDF().select("pipelinekey", "externalsourceid").distinct().count()
+    val uniqueTopicsCount = topicCounts.toDF().select("conjunctiontopics").distinct().count()
+
+    val pt = this.periodType
+    val k = ComputedAnomalyKmeansInputRecord.getK(pt, uniqueSourcesCount.toInt, uniqueTopicsCount.toInt)
+    val kmeans = new KMeans()
+    kmeans.setK(k)
+    kmeans.setMaxIter(20)
+    kmeans.setFeaturesCol("features")
+    kmeans.setPredictionCol("prediction")
```
c-w: Do we need this call given that KMeans is unsupervised?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedAnomalyKmeansModelInputDStream.scala
```diff
@@ -0,0 +1,149 @@
...
+           uniqueTopicsCount: Int): Int = {
+    var k = 31 /* max days in a month */
+    k *= uniqueSourcesCount
+    k *= uniqueTopicsCount
+    k
+  }
+
+  def features(periodType: PeriodType,
+               topicCount: TopicCount): Vector = {
+    val calendar = new GregorianCalendar()
```
c-w: The old Java Calendar API is [pretty broken](https://stackoverflow.com/q/1571265/3817588). I'd recommend against using it and use the [Java8 Joda-time API](https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html#getDayOfMonth--) instead which is safer and less verbose:

```java
ZonedDateTime.ofInstant(Instant.ofEpochMilli(topicCount.periodstartdate), ZoneId.of("UTC")).getDayOfMonth()
```

c-w (approved):
Commenting only on code, not on the overall model/approach (will sync up in person).

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedAnomalyKmeansInputRecord.scala
```diff
@@ -0,0 +1,29 @@
...
+object ComputedAnomalyKmeansInputRecord {
+
+  def apply(topicCount: TopicCount,
+            parameters: ComputedAnomalyKmeansModelParameter): ComputedAnomalyKmeansInputRecord = ComputedAnomalyKmeansInputRecord(
+    topicCount.period,
+    topicCount.periodtype,
+    topicCount.pipelinekey,
+    topicCount.externalsourceid,
+    Vectors.dense(
+      parameters.periodFeature(topicCount),
```
Smarker: do you need `.toDouble` here?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedAnomalyKmeansModelParameter.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  def kmeansParameters(periodType: PeriodType,
+                       uniqueSourcesCount: Int,
+                       uniqueTopicsCount: Int): Seq[ComputedAnomalyKmeansModelParameter] = {
+    periodType match {
+      case PeriodType.Minute => Seq(
+        ComputedAnomalyKmeansModelParameter(
+          ComputedAnomalyKmeansModelKey.Minute,
+          60 * uniqueSourcesCount * uniqueTopicsCount,
+          20,
```
Smarker: Maybe you could make the k for kmeans a constant to make it a little more clear?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTileToKmeansComputedAnomaly.scala
```diff
@@ -0,0 +1,52 @@
...
+      models(tile, session).flatMap(model => {
+        val parameters = ComputedAnomalyKmeansModelParameter.parametersForAnalysisByKey(model.key)
+        val topicCount = TopicCount(tile)
+        val input = ComputedAnomalyKmeansInputRecord(topicCount, parameters)
+        val kmeansModel = model.dataAs[KMeansModel]()
+        val tileDF = session.createDataFrame(Seq(input))
+
+        val modelCost = model.metadata("cost").toDouble
+        val computedCost = kmeansModel.computeCost(tileDF)
+        if (computedCost < computedCost) Seq()
```
Smarker: I am a little confused by this if statement's condition: `computedCost < computedCost`. I believe this would always be false.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTileToKmeansComputedAnomaly.scala
```diff
@@ -0,0 +1,52 @@
...
+import com.microsoft.partnercatalyst.fortis.spark.dto.{ComputedAnomaly, ComputedTile, MLModel, TopicCount}
+import org.apache.spark.ml.clustering.KMeansModel
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.SparkSession
+import org.joda.time.Instant
+
+object ComputedTileToKmeansComputedAnomaly {
+
+  def apply(tiles: RDD[ComputedTile], session: SparkSession): RDD[ComputedAnomaly] = {
+    tiles.flatMap(tile=>{
```
Smarker: Nit: spacing for =>

Smarker (changes requested):
I think it LGTM, but there's one if statement I feel seems off and might need to be changed before merging.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedAnomalyKmeansInputRecord.scala
jcjimenez: I don't think it's necessary - at least, in my testing, I saw Scala automatically doing the right thing, e.g.

```
scala> val a: Int = 30
a: Int = 30

scala> val b: Double = a
b: Double = 30.0
```

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTileToKmeansComputedAnomaly.scala
jcjimenez: Woah, good catch! I flipped the condition and ended up using the same variable on both sides. It should read like:

```
        if (computedCost < modelCost) Seq()
```

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedAnomalyKmeansInputRecord.scala
Smarker: ok cool

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/timeseries/ComputedTileToKmeansComputedAnomaly.scala
Smarker: oh haha I see. :)

Smarker (approved):

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/dto/TopicCount.scala
```diff
@@ -0,0 +1,31 @@
+package com.microsoft.partnercatalyst.fortis.spark.dto
+
+case class TopicCount(
+  periodtype: String,
```
noodlefrenzy, line 4: just curious on the lack of camelCasing - is that standard on the project?

jcjimenez: Yup, I've been following the pattern for the other Fortis entities.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/statistics/ComputedAnomalyKmeansModelParameter.scala
```diff
@@ -0,0 +1,112 @@
...
+  val parametersForAnalysisByKey: Map[String, ComputedAnomalyKmeansModelParameter] = parametersForAnalysis.groupBy(_.key).mapValues(v=>v.head)
+
+  def kmeansParameters(periodType: PeriodType,
+                       uniqueSourcesCount: Int,
+                       uniqueTopicsCount: Int): Seq[ComputedAnomalyKmeansModelParameter] = {
+    periodType match {
+      case PeriodType.Minute => Seq(
+        ComputedAnomalyKmeansModelParameter(
+          ComputedAnomalyKmeansModelKey.Minute,
+          60 * uniqueSourcesCount * uniqueTopicsCount, /* base of 60 buckets representing each minute within an hour */
```
noodlefrenzy, line 28: Good commenting, but might consider the ability to parameterize number of buckets to allow easier experimentation.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/statistics/ComputedAnomalyKmeansModelParameter.scala
```diff
@@ -0,0 +1,112 @@
...
+          20,
+          topicCount => {
+            100 * /* spacing between minute features */
```
noodlefrenzy, line 31: I'm a big non-fan of magic constants, so I'd either consider factoring this into a well-named function or some sort of alternate approach.

noodlefrenzy (approved):
Approach seems reasonable for what it's doing, although I'm not totally convinced that anomalies detected in this fashion are true anomalies because I'm not convinced that the data is seasonal in this way. I'd recommend experimenting and validating that detected anomalies make sense before wide deployment.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/statistics/ComputedAnomalyKmeansModelParameter.scala
jcjimenez: Good point.

jcjimenez:
Closing and doing some more research on data seasonality (or lack of).

PR #65, opened 2017-08-03
Fortis Cassandra Intregration

erikschlegel:
Taking a first crack at integrating databricks C* connector with the pipeline. Currently just pushing the `event` and `popularlocations` datasets to our cassandra to validate the approach. I'll apply the decided persistence strategy to the other tables once we're all in agreement.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -22,7 +22,7 @@ libraryDependencies ++= Seq(
   "org.apache.spark" %% "spark-core" % sparkVersion,
   "org.apache.spark" %% "spark-sql" % sparkVersion,
   "org.apache.spark" %% "spark-streaming" % sparkVersion
-).map(_ % "provided")
+)
 
 // Bundled dependencies
 libraryDependencies ++= Seq(
@@ -32,6 +32,7 @@ libraryDependencies ++= Seq(
   "com.github.catalystcode" %% "streaming-instagram" % "0.0.5",
   "com.github.catalystcode" %% "streaming-facebook" % "0.0.3",
   "com.github.catalystcode" %% "streaming-bing" % "0.0.1",
+  "com.datastax.spark" %% "spark-cassandra-connector" % "2.0.2",
   "com.github.catalystcode" %% "streaming-reddit" % "0.0.2",
   "com.github.catalystcode" % "speechtotext-websockets-java" % "0.0.7",
   "org.twitter4j" % "twitter4j-stream" % "4.0.4",
--- a/build.sbt
+++ b/build.sbt
@@ -22,7 +22,7 @@ libraryDependencies ++= Seq(
   "org.apache.spark" %% "spark-core" % sparkVersion,
   "org.apache.spark" %% "spark-sql" % sparkVersion,
   "org.apache.spark" %% "spark-streaming" % sparkVersion
-)
+).map(_ % "provided")
 
 // Bundled dependencies
 libraryDependencies ++= Seq(
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/CassandraTest.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/CassandraTest.scala
@@ -1,17 +1,14 @@
 package com.microsoft.partnercatalyst.fortis.spark
 
-  import java.util.UUID
+  import java.util.{Date, Locale, UUID}
 
   import com.microsoft.partnercatalyst.fortis.spark.dto._
-  import org.apache.spark.sql.SparkSession
   import org.apache.spark.streaming.{Seconds, StreamingContext}
   import org.apache.spark.{SparkConf, SparkContext}
 
   import scala.collection.mutable
-  import java.time.Instant.now
 
   import com.microsoft.partnercatalyst.fortis.spark.sinks.cassandra._
-  import org.apache.spark.rdd.RDD
   import org.apache.spark.streaming.dstream.DStream
 
   import scala.util.Properties.envOrElse
@@ -32,7 +29,6 @@ package com.microsoft.partnercatalyst.fortis.spark
                                   sourceurl: String,
                                   sharedLocations: List[Location] = List()
                                 ) extends Details
-
     def main(args: Array[String]): Unit = {
       val appName = this.getClass.getSimpleName
       val conf = new SparkConf()
@@ -49,8 +45,8 @@ package com.microsoft.partnercatalyst.fortis.spark
 
       val testEventsRdd = ssc.sparkContext.parallelize(Seq(TestFortisEvent(
         details = TestFortisDetails(
-            eventtime = now.getEpochSecond,
-            eventid = "1122",
+            eventtime = new Date().getTime,
+            eventid = "1132",
             sourceurl = "http://cnn.com",
             pipelinekey = "twitter",
             sharedLocations = List(),
@@ -58,32 +54,36 @@ package com.microsoft.partnercatalyst.fortis.spark
             body = "test message a new change",
             title = "twitter post" ),
         analysis = Analysis(
-          sentiments = List(.5,.5),
+          sentiments = List(.5),
           locations = List(Location(wofId = "1234", confidence = Option(1.0), latitude = Option(12.21), longitude = Option(43.1))),
           keywords = List(Tag(name = "isis", confidence = Option(1.0)), Tag(name ="car", confidence = Option(1.0))),
           genders = List(Tag(name = "male", confidence = Option(1.0)), Tag(name ="female", confidence = Option(1.0))),
           entities = List(Tag(name = "putin", confidence = Option(1.0))),
           language = Option("en")
-        ))))
+        )),
+        TestFortisEvent(
+          details = TestFortisDetails(
+            eventtime = new Date().getTime,
+            eventid = "112222",
+            sourceurl = "http://bbc.com",
+            pipelinekey = "twitter",
+            sharedLocations = List(),
+            externalsourceid = "bbc",
+            body = "This is a another test message",
+            title = "twitter post" ),
+          analysis = Analysis(
+            sentiments = List(.6),
+            locations = List(Location(wofId = "1234", confidence = Option(1.0), latitude = Option(12.21), longitude = Option(43.1))),
+            keywords = List(Tag(name = "isis", confidence = Option(1.0)), Tag(name ="car", confidence = Option(1.0)), Tag(name ="bomb", confidence = Option(1.0)), Tag(name ="fatalities", confidence = Option(1.0))),
+            genders = List(Tag(name = "male", confidence = Option(1.0)), Tag(name ="female", confidence = Option(1.0))),
+            entities = List(Tag(name = "putin", confidence = Option(1.0))),
+            language = Option("en")
+          ))))
 
       val dstream = ssc.queueStream(mutable.Queue(testEventsRdd)).asInstanceOf[DStream[FortisEvent]]
-      CassandraEventsSink(Option(dstream))
+      CassandraEventsSink(Option(dstream), ssc)
       ssc.start()
       ssc.awaitTermination()
-      /*val testList = List(EventBatchEntry(eventtime = now.getEpochSecond,
-        eventid = "1122",
-        pipelinekey = "twitter",
-        externalsourceid = "cnn",
-        computedfeatures = Features(
-          mentions = -1,
-          sentiment = Sentiment(pos_avg = 0, neg_avg = 0),
-          places = List(Place(placeid = "2134", centroidlat = 12.21, centroidlon = 43.1), Place(placeid = "213", centroidlat = 11.21, centroidlon = 43.1)),
-          gender = Gender(male_mentions = 0, female_mentions = 0),
-          keywords = List("isis", "car", "bomb"),
-          entities = List(Entities(externalsource = "", externalrefid = "", name = "putin", count = 1)))))
-      //testEventsRdd.saveToCassandra("fortis", "events", writeConf = Wri
-      val flattenedMap = testList.flatMap(CassandraEventTagsSchema(_))
-      println(flattenedMap)*/
         /*   = "twitter",
           externalsourceid = "cnn",
           batchid = batchid,
@@ -99,20 +99,7 @@ package com.microsoft.partnercatalyst.fortis.spark
           eventlangcode = "en",
           sourceurl = "http://cnn.com",
           body = "test message a new change",
-          title = "twitter post")))
+          title = "twitter post")))*/
 
-        val dstream = ssc.queueStream(mutable.Queue(testEventsRdd))
-        CassandraEventsSink(Option(dstream), session)
-        testEventsRdd.saveToCassandra("fortis", "events", writeConf = WriteConf(ifNotExists = true))
-        val addedEventsDF = session.sqlContext.read.format("org.apache.spark.sql.cassandra")
-          .options(Map("keyspace" -> "fortis", "table" -> "eventbatches"))
-          .load().select("*")
-
-        addedEventsDF.createOrReplaceTempView("eventbatches")
-        val cc = CassandraConnector(conf)
-
-        val addedEvents = session.sqlContext.sql(s"select eventid, pipelinekey from eventbatches where batchid == '$batchid'")
-        val testEventsDF = testEventsRdd.toDF
-        val dedupedEvents = testEventsDF.join(addedEvents, Seq("eventid", "pipelinekey"))*/
     }
 }
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/CassandraTest.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/CassandraTest.scala
@@ -1,173 +1,105 @@
 package com.microsoft.partnercatalyst.fortis.spark
 
-<<<<<<< HEAD
-  import java.util.{Date, Locale, UUID}
+import java.util.{Date, Locale, UUID}
 
-  import com.microsoft.partnercatalyst.fortis.spark.dto._
-=======
-  import java.util.UUID
+import com.microsoft.partnercatalyst.fortis.spark.dto._
+import org.apache.spark.streaming.{Seconds, StreamingContext}
+import org.apache.spark.{SparkConf, SparkContext}
 
-  import com.microsoft.partnercatalyst.fortis.spark.dto._
-  import org.apache.spark.sql.SparkSession
->>>>>>> Rebasing
-  import org.apache.spark.streaming.{Seconds, StreamingContext}
-  import org.apache.spark.{SparkConf, SparkContext}
+import scala.collection.mutable
 
-  import scala.collection.mutable
-<<<<<<< HEAD
+import com.microsoft.partnercatalyst.fortis.spark.sinks.cassandra._
+import org.apache.spark.streaming.dstream.DStream
 
-  import com.microsoft.partnercatalyst.fortis.spark.sinks.cassandra._
-=======
-  import java.time.Instant.now
+import scala.util.Properties.envOrElse
 
-  import com.microsoft.partnercatalyst.fortis.spark.sinks.cassandra._
-  import org.apache.spark.rdd.RDD
->>>>>>> Rebasing
-  import org.apache.spark.streaming.dstream.DStream
+object CassandraTest {
+  case class TestFortisEvent(
+                              details: Details,
+                              analysis: Analysis
+                            ) extends FortisEvent
 
-  import scala.util.Properties.envOrElse
+  case class TestFortisDetails(
+                                eventid: String,
+                                eventtime: Long,
+                                body: String,
+                                externalsourceid: String,
+                                title: String,
+                                pipelinekey: String,
+                                sourceurl: String,
+                                sharedLocations: List[Location] = List()
+                              ) extends Details
+  def main(args: Array[String]): Unit = {
+    val appName = this.getClass.getSimpleName
+    val conf = new SparkConf()
+      .setAppName(appName)
+      .set("spark.cassandra.connection.host", "13.92.93.138")
+      .setIfMissing("spark.cassandra.auth.username", envOrElse("FORTIS_CASSANDRA_USER", "cassandra"))
+      .setIfMissing("spark.cassandra.auth.password", envOrElse("FORTIS_CASSANDRA_PASSWORD", "cassandra"))
+      .set("spark.cassandra.input.consistency.level", "LOCAL_ONE")
+      .set("spark.cassandra.output.consistency.level", "LOCAL_ONE")
+      .set("output.consistency.level", "LOCAL_ONE")
+      .setIfMissing("spark.master", "local[*]")
+    val ssc = new StreamingContext(conf, Seconds(10))
+    val batchid = UUID.randomUUID().toString
 
-  object CassandraTest {
-    case class TestFortisEvent(
-                                details: Details,
-                                analysis: Analysis
-                              ) extends FortisEvent
-
-    case class TestFortisDetails(
-                                  eventid: String,
-                                  eventtime: Long,
-                                  body: String,
-                                  externalsourceid: String,
-                                  title: String,
-                                  pipelinekey: String,
-                                  sourceurl: String,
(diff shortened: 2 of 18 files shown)


build.sbt
```diff
@@ -22,7 +22,7 @@ libraryDependencies ++= Seq(
   "org.apache.spark" %% "spark-core" % sparkVersion,
   "org.apache.spark" %% "spark-sql" % sparkVersion,
   "org.apache.spark" %% "spark-streaming" % sparkVersion
-).map(_ % "provided")
```
kevinhartman: The reason for this line is to avoid having these built into the fat JAR since they're available in every Spark runtime environment. Was this change intentional?

kevinhartman: Looks like in a previous change `"com.datastax.spark" %% "spark-cassandra-connector" % "2.0.2"` was added on line 21 within this `Seq`. This should not be here. Can you please remove it? Your change on line 35 in this PR will be sufficient.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/CassandraTest.scala
```diff
@@ -0,0 +1,105 @@
...
+import scala.collection.mutable
+
+import com.microsoft.partnercatalyst.fortis.spark.sinks.cassandra._
+import org.apache.spark.streaming.dstream.DStream
+
+import scala.util.Properties.envOrElse
+
+object CassandraTest {
+  case class TestFortisEvent(
+                              details: Details,
```
kevinhartman, line 18: Nesting should be a 2 space indent.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/CassandraTest.scala
```diff
@@ -0,0 +1,105 @@
...
+      .setIfMissing("spark.cassandra.auth.username", envOrElse("FORTIS_CASSANDRA_USER", "cassandra"))
+      .setIfMissing("spark.cassandra.auth.password", envOrElse("FORTIS_CASSANDRA_PASSWORD", "cassandra"))
+      .set("spark.cassandra.input.consistency.level", "LOCAL_ONE")
+      .set("spark.cassandra.output.consistency.level", "LOCAL_ONE")
+      .set("output.consistency.level", "LOCAL_ONE")
+      .setIfMissing("spark.master", "local[*]")
+    val ssc = new StreamingContext(conf, Seconds(10))
+    val batchid = UUID.randomUUID().toString
+
+    val testEventsRdd = ssc.sparkContext.parallelize(Seq(TestFortisEvent(
```
kevinhartman: nit: `TestFortisEvent` should be on next line for alignment with subsequent elements

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventSchema.scala
```diff
@@ -0,0 +1,133 @@
...
+  def apply(item: FortisEvent, batchid: String): Event = {
+    Event(
+      pipelinekey = item.details.pipelinekey,
+      externalsourceid = item.details.externalsourceid,
+      computedfeatures = Utils.getFeature(item),
+      eventtime = item.details.eventtime,
+      batchid = batchid,
+      eventlangcode = item.analysis.language.orNull,
+      eventid = item.details.eventid,
+      insertiontime = new Date().getTime,
```
kevinhartman: If this is `insertion_time` in our schema, then this should be camelcased: `insertionTime`.

build.sbt
```diff
@@ -22,7 +22,7 @@ libraryDependencies ++= Seq(
...
+)
```
jcjimenez: I'm not sure about this, but should we leave the `"provided"` part in this so as to not include these in our fat jar?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventSchema.scala
```diff
@@ -0,0 +1,133 @@
...
+
+          (sortedCombo(0), sortedCombo(1), sortedCombo(2))
+        })
+      case None => Seq()
+    }
+  }
+
+  def getStringOption(str: String): Option[String] = {
+    for (v <- Option(str) if str.nonEmpty) yield v
+  }
```
jcjimenez: Is this method called somewhere in a file not included in this PR? (I don't see any references to it)

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventSchema.scala
```diff
@@ -0,0 +1,133 @@
...
+      eventid = item.eventid,
+      topic = kw.toLowerCase,
+      eventtime = item.eventtime.getTime,
+      externalsourceid = item.externalsourceid,
+      placeid = location.placeid
+    )
+  }
+}
+
+object Utils {
```
jcjimenez, line 118: Should we make this object private to this file?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/dto/AggregationRecord.scala
```diff
@@ -0,0 +1,12 @@
+package com.microsoft.partnercatalyst.fortis.spark.sinks.cassandra.dto
+
+trait AggregationRecord {
+  val periodstartdate: Long
+  val periodenddate: Long
```
jcjimenez, line 5: I've been wondering about `periodenddate` lately: do we use it anywhere?

jcjimenez (approved):
LGTM with some questions.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventSchema.scala
erikschlegel: I considered that but I have a couple of unit tests to validate the functionality. How have we approached a similar situation in the past with the project?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/dto/AggregationRecord.scala
erikschlegel: we use it for the aggregation. I feel like it will serve its existence for the time range filters in the histogram.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventSchema.scala
jcjimenez: Ah, if your tests are in the same package, declaring them package private could be an option, e.g.

```
private[cassandra] object Utils {
```

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.spark.streaming.StreamingContext
+
+object CassandraEventsSink {
+  private val KeyspaceName = "fortis"
+  private val TableEvent = "events"
+  private val TableEventTags = "eventtags"
+  private val TableEventBatches = "eventbatches"
+  private val CassandraFormat = "org.apache.spark.sql.cassandra"
+
+  def apply(dstream: Option[DStream[FortisEvent]], ssc: StreamingContext): Unit = {
```
kevinhartman: It'd be cleaner to remove the `Option` wrapper from this parameter.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
```diff
@@ -0,0 +1,102 @@
...
+      val batchStream = dstream.get.foreachRDD(eventsRDD => {
+        eventsRDD.foreachPartition(partition => {
+          val sparkContext = eventsRDD.sparkContext
+          val session = SparkSession.builder().config(sparkContext.getConf)
+            .appName(sparkContext.appName)
+            .getOrCreate()
+          val batchid = UUID.randomUUID().toString
+          registerUDFs(session)
+          writeEventRddToEventsTable(eventsRDD, batchid)
+          val eventBatchDF = fecthEventBatch(batchid, session)
```
kevinhartman: If there's a failure here or later prior to writing the aggregation(s) out, we will orphan events, right? We will never visit their batch ID, and the corresponding events, if repeated from the source stream, will not re-persist since they will already exist in the events table with the orphaned batch ID.

Some thought may be required here. Ideally, we should get the events of any batch IDs that aggregation hasn't run for yet. As a quick fix, maybe we could add a table recording batch ID and processing status (aggregation completed, or not), and select all events where the corresponding batch has not yet been aggregated. Timing considerations should be thought through with this kind of approach as it may be possible for `foreachRdd` to execute in parallel.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
```diff
@@ -0,0 +1,102 @@
...
+  private val TableEventTags = "eventtags"
+  private val TableEventBatches = "eventbatches"
+  private val CassandraFormat = "org.apache.spark.sql.cassandra"
+
+  def apply(dstream: Option[DStream[FortisEvent]], ssc: StreamingContext): Unit = {
+    if (dstream.isDefined) {
+
+      val aggregators = Seq(new PopularPlacesAggregator)
+      val batchStream = dstream.get.foreachRDD(eventsRDD => {
+        eventsRDD.foreachPartition(partition => {
```
kevinhartman: This line must be removed. Otherwise we're asking Spark to serialize `eventsRDD` to worker nodes  (since it's used inside this block), and then we're trying to set up additional RDD transformations on those worker nodes (this doesn't make sense since RDD graphs are built on driver only).

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
kevinhartman: Additionally, this would violate the expectation that all events returned would exist in the current RDD, so that may prevent the optimization of only fetching event IDs.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
kevinhartman: To clarify, no further change is required other than removing this line.

kevinhartman (changes requested):
I've added a few preliminary comments regarding the approach we're taking to connect aggregation to ingestion. Some of the issues raised I expect may result in code churn within this PR, so I'm holding off on a thorough review until they've been addressed.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
erikschlegel: This was a mistake on my end as I forgot to remove the `forEachPartition` call from an alternative approach I attempted earlier. I'll remove.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventSchema.scala
```diff
@@ -0,0 +1,128 @@
...
+      pipelinekey = item.pipelinekey,
+      centroidlat = location.centroidlat,
+      centroidlon = location.centroidlon,
+      placeid = location.placeid,
+      periodstartdate = Period(item.eventtime.getTime, periodType).startTime(),
+      periodenddate = Period(item.eventtime.getTime, periodType).endTime(),
+      periodtype = periodType.periodTypeName,
+      period = periodType.format(item.eventtime.getTime),
+      externalsourceid = item.externalsourceid,
+      mentioncount = item.computedfeatures.mentions,
```
c-w: Summarizing conversation over the phone. We can simplify this by using the `counter` column type.

This has a number of advantages, most prominently:
1) We don't have to read the current value for every write.
2) We avoid race conditions as Cassandra counters are guaranteed to be eventually consistent.
3) The code will be simpler as aggregations are now just single writes.

In essence I propose a schema like this:

```cql
CREATE TABLE myAggregation_data (
  key_col1 text,
  key_col2 text,
  data_col1 text,
  counter_pointer text,
  PRIMARY KEY (key_col1, key_col2)
);

-- need this separate table as Cassandra doesn't allow
-- counter columns on a table that has non-key columns
-- so the following wouldn't work:
-- CREATE TABLE foo (
--   key_col text,
--   data_col text,
--   counter_col counter,
--   PRIMARY KEY (key_col)
-- );
CREATE TABLE myAggregation_counters (
  id text,
  value1 counter,
  value2 counter,
  length counter,
  PRIMARY KEY (id)
);
```

We can insert into this schema like so:

```cql
INSERT INTO myAggregation_data VALUES (
  -- same as we insert now
) IF NOT EXISTS;

UPDATE myAggregation_counters
SET value1 = value1 + 23
  , value2 = value2 + 12
  , length = length + 1
WHERE id = ?;
```

We can then read back the average values like so:

```js
const dataQuery = `
// same query as now, only excluding the mentioncount/sentiment_avg columns
`.trim();

const dataParams = [
  // from the graphql args, as we query now
];

const counterQuery = `
SELECT * FROM myAggregation_counters
WHERE id IN ?
`.trim();

cassandraConnector.execute(dataQuery, dataParams)
.then(rows => {
  const counterIds = rows.map(row => row.counter_pointer);
  cassandraConnector.execute(counterQuery, [counterIds])
  .then(counters => {
    const avg_value1 = mean(counters.map(counter =>
      counter.value1 / counter.length));

    // do something with avg_value1 here
  });
});
```

Limitations of this approach:
- For each read, we make two trips to Cassandra -- this is OK because we have many more writes than reads.
- Limited precision as we can no longer store floats, counters are just integers. This is okay because we can just truncate-multiply the sentiment float value like 0.754 => 754.

c-w (dismissed):
We figured out a way to make this simpler :)

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventSchema.scala
```diff
@@ -0,0 +1,173 @@
...
+      periodstartdate = Period(item.eventtime, periodType).startTime(),
+      periodenddate = Period(item.eventtime, periodType).endTime(),
+      periodtype = periodType.periodTypeName,
+      period = periodType.format(item.eventtime),
+      externalsourceid = item.externalsourceid,
+      mentioncount = item.computedfeatures.mentions,
+      conjunctiontopic1 = kw._1,
+      conjunctiontopic2 = kw._2,
+      conjunctiontopic3 = kw._3,
+      avgsentimentnumerator = 0,
```
c-w, line 54: Given that we already have one sentiment in the numerator, should the denominator be 1?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventSchema.scala
```diff
@@ -0,0 +1,173 @@
...
+        })
+      case None => Seq()
+    }
+  }
+
+  def getSentimentScore(sentiments: List[Double]): Float = {
+    Option(sentiments) match {
+      case None => 0F
+      case Some(sentimentList) => {
+        var neg_sent = 0F
```
c-w, line 144: This may be clearer with a ternary as it'd avoid the unidiomatic var:

```scala
if (sentimentList.isEmpty) 0F else sentimentList.head.toFloat
```

That being said: why the need to cast this to float instead of changing the sentiment field in the database to a double? That would avoid us loosing precision.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventSchema.scala
```diff
@@ -0,0 +1,173 @@
...
+          neg_sent = sentimentList.head.toFloat
+        }
+
+        neg_sent
+      }
+    }
+  }
+
+  def getFeature(item: FortisEvent): Features = {
+    val genderCounts = item.analysis.genders.map(_.name).groupBy(identity).mapValues(t=>t.size.toLong)
```
c-w, line 155: Given that we don't have gender as a feature yet, shall we make this explicit by not implementing the aggregation?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
```diff
@@ -0,0 +1,138 @@
...
+        if(!eventsRDD.isEmpty) {
+          val batchid = UUID.randomUUID().toString
+          val fortisEvents = writeFortisEvents(eventsRDD, batchid)
+          val aggregators = Seq(new PopularPlacesAggregator, new PopularTopicAggregator).par
+          val session = SparkSession.builder().config(eventsRDD.sparkContext.getConf)
+            .appName(eventsRDD.sparkContext.appName)
+            .getOrCreate()
+
+          if(session != null){
+            registerUDFs(session)
```
c-w, line 40: Should this only happen once, at app startup?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
```diff
@@ -0,0 +1,138 @@
...
+
+  val eventTopicsRddColumns = SomeColumns("topic", "pipelinekey", "externalsourceid", "insertiontime", "eventids" append)
+  val eventPlacesRddColumns = SomeColumns("conjunctiontopic1", "conjunctiontopic2","conjunctiontopic3", "insertiontime", "centroidlat", "centroidlon", "pipelinekey", "placeid", "externalsourceid", "eventids" append)
+  //Need to convert the DS to a RDD as C* column append isnt support for Dataframes
+  saveRddToCassandra[EventTopics](KeyspaceName, TableEventTopics, eventDS.flatMap(CassandraEventTopicSchema(_)).rdd, eventTopicsRddColumns)
+  saveRddToCassandra[EventPlaces](KeyspaceName, TableEventPlaces, eventDS.flatMap(CassandraEventPlacesSchema(_)).rdd, eventPlacesRddColumns)
+}
+
+private def saveRddToCassandra[T](keyspace: String, table: String, rdd: RDD[T], columns: ColumnSelector, attempt: Int = 0)(
+                                 // implicit parameters as a separate list!
```
c-w, line 98: Would it be possible to make this comment redundant by reformatting like this?

```scala
implicit
rwf: RowWriterFactory[T],
columnMapper: ColumnMapper[T]
```

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
```diff
@@ -0,0 +1,138 @@
...
+                                 implicit rwf: RowWriterFactory[T],
+                                 columnMapper: ColumnMapper[T]
+                               ): Unit = {
+Try(rdd.saveToCassandra(keyspace, table, columns)) match {
+  case Success(_) => None
+  case Failure(ex) =>
+    ex.printStackTrace
```
c-w, line 105: This should probably be a log, not a print.

Also, as a general note: in Scala we often use the parentheses when calling a no-argument stateful method like print, so: `ex.printStackTrace()`, not `ex.printStackTrace`.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
```diff
@@ -0,0 +1,138 @@
...
+    Thread.sleep(CassandraRetryIntervalMS)
```
c-w, line 106: I wonder if you want this as an exponential backoff instead of a linear one: `CassandraRetryIntervalMS * attempt` if we initialize the attempt to 1 instead of zero on line 116 below.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
```diff
@@ -0,0 +1,138 @@
...
+
+  val incrementallyUpdatedDF = aggregator.IncrementalUpdate(session, aggregatedDF)
+  incrementallyUpdatedDF.write
+    .format(CassandraFormat)
+    .mode(SaveMode.Append)
+    .options(Map("keyspace" -> KeyspaceName, "table" -> aggregator.FortisTargetTablename)).save
+} catch {
+  case ex if attempt < CassandraMaxRetryAttempts => {
+    ex.printStackTrace
+    aggregateEventBatch(eventDS, session, aggregator, attempt + 1)
```
c-w, line 132: Is this missing a `Thread.sleep` like above?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
```diff
@@ -0,0 +1,138 @@
...
+          val aggregators = Seq(new PopularPlacesAggregator, new PopularTopicAggregator).par
+          val session = SparkSession.builder().config(eventsRDD.sparkContext.getConf)
+            .appName(eventsRDD.sparkContext.appName)
+            .getOrCreate()
+
+          if(session != null){
+            registerUDFs(session)
+            val eventBatchDF = fetchEventBatch(batchid, fortisEvents, session)
+            writeEventBatchToEventTagTables(eventBatchDF, session)
+            aggregators.map(aggregator => {
```
c-w, line 43: Hm, still not sure we really need the in-Spark aggregation as opposed to just updating Cassandra with rows.

How much data do we expect to have in any specific batch?

If it's less than 65k, we can use Cassandra's [batch keyword](http://docs.datastax.com/en/cql/3.1/cql/cql_reference/batch_r.html) to simplify the insertion logic but just building up a bunch of simple insert statements for the RDD and then inserting them in one go.

If we can't afford one insert statement per 65k records, we can use the [copy statement](http://docs.datastax.com/en/cql/3.1/cql/cql_reference/copy_r.html) and stream the RDD rows to Cassandra.

Wondering what your thoughts are for these.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
```diff
@@ -0,0 +1,138 @@
...
+
+  private def fetchEventBatch(batchid: String, events: RDD[Event], session: SparkSession): Dataset[Event] = {
+    import session.sqlContext.implicits._
+
+    val addedEventsDF = session.sqlContext.read.format(CassandraFormat)
+      .options(Map("keyspace" -> KeyspaceName, "table" -> TableEventBatches))
+      .load()
+
+    addedEventsDF.createOrReplaceTempView(TableEventBatches)
+    val ds = session.sqlContext.sql(s"select eventid, pipelinekey " +
```
c-w, line 77: NB: you could just use a multiline string here to save yourself all those pluses and quote characters:

```scala
val ds = session.sqlContext.sql(s"""select eventid, pipelinekey
                                  | from ${TableEventBatches}
                                  | where batchid = '${batchid}'""".stripMargin)
```

PR #66, opened 2017-08-04
DBSCAN clustering POC

jcjimenez:
DO NOT MERGE - for review purposes only

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -20,6 +20,7 @@ parallelExecution in Test := false
 libraryDependencies ++= Seq(
   "com.datastax.spark" %% "spark-cassandra-connector" % "2.0.2",
   "org.apache.spark" %% "spark-core" % sparkVersion,
+  "org.apache.spark" %% "spark-mllib" % sparkVersion,
   "org.apache.spark" %% "spark-sql" % sparkVersion,
   "org.apache.spark" %% "spark-streaming" % sparkVersion
 ).map(_ % "provided")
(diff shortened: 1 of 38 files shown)


build.sbt
```diff
@@ -49,6 +53,8 @@ libraryDependencies ++= Seq(
   "eus.ixa" % "ixa-pipe-nerc" % "1.6.1"
 )
 
+libraryDependencies += "nz.ac.waikato.cms.weka" % "optics_dbScan" % "1.0.4"
```
c-w, line 56: Oh man, what a blast from the past. I remember Weka from college ;) Would't recommend using it for any sort of non-toy data as their implementations are not well optimized: it's a teaching tool first and foremost, not a production tool.

As an alternative: I've heard good things about [ELKI](https://elki-project.github.io/releases/current/doc/de/lmu/ifi/dbs/elki/algorithm/clustering/DBSCAN.html).

c-w, line 56: You made some visualizations of the fortis-v1 data, right? I remember your post about plotly. Could you share the visualizations with the group so that we can brainstorm about what sorts of assumptions we can and cannot make about the data?

c-w (changes requested):

build.sbt
jcjimenez: Yeah, I was trying this to compare to dbscan on spark - both of which may not be ready for prime time for our use.

c-w: DBSCAN is not trivially parallizable so I'm not surprised that the in-proc implementations rule for now :)

PR #67, opened 2017-08-04
Implement Cassandra configuration manager

kevinhartman:
Implementation of the `ConfigurationManager` interface for Cassandra.

**Summary**
- Cassandra config manager implementation. Uses Spark to perform required manipulations.
- Adds Spark context to configuration manager parameter lists for DB access backed by Spark.
- Changes `ConnectorConfig.parameters` from `Map[String, String]` to `Map[String, Any]`.
- Adds `ParameterExtensions` helper class to simplify reading objects from connector config parameter maps.

### DIFF
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/dba/CassandraConfigurationManager.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/dba/CassandraConfigurationManager.scala
@@ -1,0 +1,84 @@
+package com.microsoft.partnercatalyst.fortis.spark.dba
+import java.util.concurrent.ConcurrentHashMap
+
+import com.microsoft.partnercatalyst.fortis.spark.dto.{BlacklistedTerm, Geofence, SiteSettings}
+import com.microsoft.partnercatalyst.fortis.spark.sources.streamprovider.ConnectorConfig
+import org.apache.spark.SparkContext
+import com.datastax.spark.connector._
+import com.microsoft.partnercatalyst.fortis.spark.logging.Loggable
+import scala.compat.java8.FunctionConverters._
+
+@SerialVersionUID(100L)
+class CassandraConfigurationManager extends ConfigurationManager with Serializable with Loggable {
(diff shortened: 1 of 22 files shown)


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/dba/CassandraConfigurationManager.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  override def fetchConnectorConfigs(sparkContext: SparkContext, pipeline: String): List[ConnectorConfig] = {
+    def fetchTrustedSources(connectorName: String): Seq[String] = {
+      sparkContext.cassandraTable(CassandraSchema.KeyspaceName, CassandraSchema.Table.TrustedSourcesName)
+        .select("externalsourceid")
+        .map(row => row.getString("externalsourceid")).collect()
+    }
+
+    val pipelineConfigRows = sparkContext.cassandraTable[CassandraSchema.Table.Stream](CassandraSchema.KeyspaceName,
+      CassandraSchema.Table.StreamsName).collect()
```
jcjimenez, line 25: No biggie, but I think we could get rid of (or defer) this `collect()` call since we call `toList` at the end - this way we only iterate through the list once.

jcjimenez (approved):
LGTM

Smarker (approved):
LGTM

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/dba/CassandraConfigurationManager.scala
kevinhartman: Hey so what I'm doing here is bringing the stream row objects from the worker nodes to the driver. The reason for this is so that the map call below it will execute on the driver (I do more Cassandra queries in there, and also access a local cache on the driver).

jcjimenez: Ah, just noticed the `computeIfAbsent`!

PR #75, opened 2017-08-09
Fix issue with getting streams for pipeline

kevinhartman:
Previously, all configs in `streams` table were being returned for all pipelines due to missing `where` clause.


jcjimenez (approved):
LGTM

PR #77, opened 2017-08-11
Guarding against null data attribute of RedditObject instance

jcjimenez:
The null data may happen for certain types of RedditObject.


jcjimenez:
Cosing in favor of an approach based on `requiredValuesProvided`

PR #83, opened 2017-08-11
Implemented ConjunctiveTopicsAggregator.

jcjimenez:
screen_shot_2017-08-11_at_4_51_18_pm


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventSchema.scala
```diff
@@ -79,6 +79,37 @@ object CassandraPopularTopics {
   }
 }
 
+object CassandraConjunctiveTopics {
+  def apply(item: Event): Seq[ConjunctiveTopicAggregate] = {
+    val keywords = item.computedfeatures.keywords
+    val keywordPairs = keywords.combinations(2).flatMap(combination=>Seq(
```
erikschlegel: What happens if there's only a single keyword in the message?

erikschlegel (approved):
LGTM and great work. Can you please opine to the posted question.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventSchema.scala
jcjimenez: If there's a single keyword, the `combinations(2)` call returns an empty sequence. So, the combination in the `for` loop below short-circuits and this method emits an empty sequence.

erikschlegel: That's what I worried about. If there's only one term hit then there should be one record in this table with the total mention count for that term.

jcjimenez: got it! should the `conjunctivetopic` property be set to null or the same value as `topic`?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventSchema.scala
```diff
@@ -81,6 +81,37 @@ object CassandraPopularTopics {
   }
 }
 
+object CassandraConjunctiveTopics {
+  def apply(item: Event): Seq[ConjunctiveTopicAggregate] = {
+    val keywords = item.computedfeatures.keywords
+    val keywordPairs = if (keywords.size == 1) Seq((keywords.head, null)) else keywords.combinations(2).flatMap(combination => Seq(
```
jcjimenez, line 87: Hi @erikschlegel , what do you think?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
```diff
@@ -31,7 +31,12 @@ object CassandraEventsSink{
           val batchid = UUID.randomUUID().toString
           val fortisEventsRDD = eventsRDD.map(CassandraEventSchema(_, batchid))
           writeFortisEvents(fortisEventsRDD, batchid)
-          val aggregators = Seq(new PopularPlacesAggregator, new PopularTopicAggregator, new ComputedTilesAggregator).par
+          val aggregators = Seq(
+            new ConjunctiveTopicsAggregator,
+            new PopularPlacesAggregator,
+            new PopularTopicAggregator,
+            new ComputedTilesAggregator
```
jcjimenez: Putting these in different lines to avoid merge conflicts.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/dto/FortisRecords.scala
```diff
@@ -99,6 +99,21 @@ case class PopularTopicAggregate(
                                   topic: String
                                 ) extends AggregationRecordTile with Serializable
 
+case class ConjunctiveTopicAggregate(
+                                      conjunctivetopic: String,
```
kevinhartman, line 103: Hey can you fix the indentation in this file to follow our convention of 2 space indents while you're here? :)

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventSchema.scala
Smarker, line 87: `keywords` won't ever be null here right?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/dto/FortisRecords.scala
jcjimenez: Sure, mind if I put this in a different PR?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventSchema.scala
jcjimenez: Correct, based on other similar references (e.g. `CassandraEventTopicSchema#apply`), this  should be safe.

PR #90, opened 2017-08-15
Added filter against *.orig

jcjimenez:
Super simple.


c-w (approved):
Is .orig some sort of backup file on Mac?

.gitignore
```diff
@@ -5,6 +5,8 @@
 # macOS
 .DS_Store
 
+**/*.orig
```
Smarker: What about just using `*.orig`?

jcjimenez:
@c-w it's actually a file git generates when something goes wrong during merging - I assume it's to help one figure out the before/after of a particular file.

.gitignore
jcjimenez: `*.orig` will capture files at the base of the project directory, so adding `**/` in front of that will ensure that any files no matter their parent directory should be ignored (e.g. src/main/java/com/some/package/SomeClass.scala.orig)

PR #92, opened 2017-08-16
Replace Cognitive Services language detector

c-w:
As per conversation yesterday: we want to reduce the number of cognitive services calls so replacing Cognitive Services language detector with a local one.


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/logging/AppInsightsTelemetry.scala
```diff
@@ -1,29 +1,40 @@
...
+
 import com.microsoft.applicationinsights.{TelemetryClient, TelemetryConfiguration}
-import java.util.HashMap
 
 class AppInsightsTelemetry extends FortisTelemetry {
   private val client: TelemetryClient = new TelemetryClient(TelemetryConfiguration.createDefault())
 
   def logIncomingEventBatch(streamId: String, connectorName: String, batchSize: Long): Unit = {
-    val properties = new HashMap[String, String](2)
+    val properties = new util.HashMap[String, String](2)
```
Smarker, line 11: Why would using `util.HashMap` be better than using `HashMap` here?

c-w: This is to fix a Scala warning. More details [here](https://stackoverflow.com/q/18196520/3817588) or [here](https://blog.jetbrains.com/scala/2012/05/26/mutableimmutable-collections-prefix-convention-support/). Apparently it's a convention in the Scala community to make non-standard collections (e.g. java.util, scala.mutable) stand out more.

Mostly I just didn't want the bright yellow warning highlights on my screen since they're distracting from actual issues and these ones were very easy to fix :P

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/language/LocalLanguageDetector.scala
```diff
@@ -0,0 +1,24 @@
...
+import com.optimaize.langdetect.LanguageDetectorBuilder
+import com.optimaize.langdetect.ngram.NgramExtractors
+import com.optimaize.langdetect.profiles.LanguageProfileReader
+import com.optimaize.langdetect.text.CommonTextObjectFactories
+
+@SerialVersionUID(100L)
+class LocalLanguageDetector extends LanguageDetector {
+  @transient private lazy val languageProfiles = new LanguageProfileReader().readAllBuiltIn
+  @transient private lazy val languageDetector = LanguageDetectorBuilder.create(NgramExtractors.standard()).withProfiles(languageProfiles).build()
+  @transient private lazy val textObjectFactory = CommonTextObjectFactories.forDetectingOnLargeText()
```
kevinhartman: Is large text always the best choice here? I see the other relevant-sounding option is `forDetectingShortCleanText`. While Tweets probably aren't clean, they're also not long.

jcjimenez (approved):
LGTM

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Pipeline.scala
```diff
@@ -57,7 +57,7 @@ object Pipeline {
       }
 
       def addLanguage(event: ExtendedFortisEvent[T]): ExtendedFortisEvent[T] = {
-        val language = analyzer.detectLanguage(event.details, languageDetector)
+        val language = analyzer.detectLanguage(event.details, new LocalLanguageDetector)
```
kevinhartman: Since this is on a per-event code path, it'll get constructed for every event, so the lazy vals in the impl won't be reused. Are they light-weight to construct?

c-w: Moved outside in ffc18e8

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/language/LocalLanguageDetector.scala
c-w: We don't only have Tweets but also news text, Facebook posts, comments, etc. so overall the default detector is probably fine. We can also run it through both detectors to potentially increase accuracy.

c-w: Done in b0fff7d.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/logging/AppInsightsTelemetry.scala
```diff
@@ -1,29 +1,40 @@
...
 
-    val metrics = new HashMap[String, java.lang.Double](2)
+    val metrics = new util.HashMap[String, java.lang.Double](2)
     metrics.put("batchSize", batchSize.toDouble)
     metrics.put("duration", duration.toDouble)
 
     client.trackEvent("batch.sink", properties, metrics)
   }
+
+  def logLanguageDetection(language: Option[String]): Unit = {
```
erikschlegel, line 31: Isn't this a general utility class for writing events into app insights?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/language/CognitiveServicesLanguageDetector.scala
```diff
@@ -0,0 +1,56 @@
...
+      return None
+    }
+
+    val textId = "0"
+    val requestBody = buildRequestBody(text, textId)
+    val response = callCognitiveServices(requestBody)
+    parseResponse(response, textId)
+  }
+
+  protected def callCognitiveServices(requestBody: String): String = {
```
erikschlegel, line 26: Why would we still need to call the cog svc language endpoint?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/language/LocalLanguageDetector.scala
c-w: Actually, we can based it off of text length af093c6

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/language/CognitiveServicesLanguageDetector.scala
c-w: This is the old class that got renamed. I can delete it if you want, but given that we're keeping the Kafka sink around too, might as well keep this option around.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/logging/AppInsightsTelemetry.scala
c-w: This is the class that's being used to log events to AppInsights, so that's I reckon where additional event logging should be encapsulated.

kevinhartman (approved):

PR #93, opened 2017-08-16
Fix a few IDEA inspections

c-w:
No functional change.


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transformcontext/TransformContextProvider.scala
```diff
@@ -21,8 +21,8 @@ class TransformContextProvider(configManager: ConfigurationManager, featureServi
 
   // Do not serialize these values. The transformContext would otherwise contain stale Broadcast
   // instances, so we rebuild it on recovery from checkpoint.
-  @volatile @transient private var transformContext: TransformContext = null
-  @volatile @transient private var queueClient: QueueClient = null
+  @volatile @transient private var transformContext: TransformContext = _
```
kevinhartman, line 24: IMO, `null` is more explicit, since we're expecting this to be exactly `null` (i.e. line 48), though I understand this likely fixes a style warning.

kevinhartman (approved):

PR #100, opened 2017-08-21
Some minor fixes to the computed tiles aggregator

erikschlegel:


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/CassandraTest.scala
```diff
@@ -35,7 +35,7 @@ object CassandraTest {
     val appName = this.getClass.getSimpleName
     val conf = new SparkConf()
       .setAppName(appName)
-      .set("spark.cassandra.connection.host", "52.177.191.215")
+      .set("spark.cassandra.connection.host", "40.79.72.103")
```
c-w: Should this come from an environment variable so we don't have to re-compile the binary every time this changes? We already have `FORTIS_CASSANDRA_HOST` ([link](https://github.com/CatalystCode/project-fortis-pipeline/blob/32e8c1795a787db6ce4ae181e746d4d754f4ae19/ops/setup-environment.sh#L30)) so maybe we should just read this here?

c-w (approved):

kevinhartman (approved):

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/udfs/WeightedAvg.scala
```diff
@@ -44,6 +44,10 @@ abstract class WeightedAvg extends UserDefinedAggregateFunction {
 }
 
 object SentimentWeightedAvg extends WeightedAvg with Serializable{
+  override def evaluate(buffer: Row): Double = {
+    getDoubleValue(Option(buffer.getDouble(0))) / getLongValue(Option(buffer.getLong(1))).toDouble
```
Smarker, line 48: I believe this would throw an exception if `getLongValue(Option(buffer.getLong(1))).toDouble` returns 0. I would put a guard against this just to protect against divide by zero.

Smarker (approved):
LGTM

PR #102, opened 2017-08-22
Proposal for streams changes listener.

jcjimenez:
empty_stream


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
```diff
@@ -61,8 +63,13 @@ object ProjectFortis extends App {
...
 
-    val sparksession = SparkSession.builder().config(conf).getOrCreate()
-    val ssc = new StreamingContext(sparksession.sparkContext, batchDuration)
+    val sparkContext = SparkContext.getOrCreate(conf)
+    val ssc = new StreamingContext(sparkContext, batchDuration)
+    ssc.checkpoint(fortisSettings.progressDir)
+    ssc
+  }
+
+  private def attachToContext(ssc:StreamingContext): Boolean = {
```
jcjimenez, line 71: I factored out `attachToContext` from `createStreamingContext` in order to call it in the loop below.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/StreamsChangeListener.scala
```diff
@@ -0,0 +1,65 @@
...
+        logInfo(s"Service Bus message for updated streams received; Re-scheduling streaming context stop for ${contextStopWaitTimeMillis} milliseconds from now.")
+        this.timer.get.cancel()
+      } else {
+        logInfo(s"Service Bus message for updated streams received; Requesting streaming context stop in ${contextStopWaitTimeMillis} milliseconds.")
+      }
+
+      val t = new Timer()
+      t.schedule(new TimerTask {
+        override def run(): Unit = {
+          ssc.stop(stopSparkContext = true, stopGracefully = true)
```
jcjimenez: Not sure if just calling `stop` is enough to dispose of the open streams.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Constants.scala
```diff
@@ -6,6 +6,8 @@ object Constants {
   val SparkStreamingBatchSizeDefault = 5
 
   val EventHubProgressDir = "eventhubProgress"
+  val ContextStopWaitTimeMillis = 5000
```
jcjimenez: I think I need to change this to allow for more wiggle room in cases where the user is adding more than one stream. Maybe increase this to 1 minute?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Constants.scala
```diff
@@ -18,5 +20,7 @@ object Constants {
     val ManagementBusConnectionString = "FORTIS_SB_CONN_STR"
     val ManagementBusConfigQueueName = "FORTIS_SB_CONFIG_QUEUE"
     val ManagementBusCommandQueueName = "FORTIS_SB_COMMAND_QUEUE"
+    val ContextStopWaitTimeMillis = "FORTIS_CONTEXT_STOP_WAIT_TIME_MILLIS"
```
kevinhartman: nit: Can you make these name a little shorter?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
```diff
@@ -85,14 +92,28 @@ object ProjectFortis extends App {
...
+    true
   }
 
   // Main starts here
-  val ssc = StreamingContext.getOrCreate(fortisSettings.progressDir, createStreamingContext)
+  while(true) {
+    logInfo("Creating streaming context.")
+    val ssc = StreamingContext.getOrCreate(fortisSettings.progressDir, createStreamingContext)
+    var didAttach = attachToContext(ssc)
+    while (!didAttach) {
```
kevinhartman: Do you need to store `didAttach`?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/StreamsChangeListener.scala
```diff
@@ -0,0 +1,66 @@
...
+import java.util.{Timer, TimerTask}
+
+import com.microsoft.azure.servicebus._
+import com.microsoft.azure.servicebus.primitives.ConnectionStringBuilder
+import com.microsoft.partnercatalyst.fortis.spark.logging.Loggable
+import org.apache.spark.streaming.StreamingContext
+
+object StreamsChangeListener {
+
+  @volatile @transient private var queueClient: QueueClient = _
```
kevinhartman: This object won't be serialized so you should be able to make this a `val` right? In that way, it should also be safe to access from the inner class helper thread.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/StreamsChangeListener.scala
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    override def onMessageAsync(message: IMessage): CompletableFuture[Void] = {
+      if (message.getLabel != "streamsDidChange") {
+        return CompletableFuture.completedFuture(null)
+      }
+
+      val contextStopWaitTimeMillis = settings.contextStopWaitTimeMillis
+
+      if (this.timer.isDefined) {
```
kevinhartman: How about a `match` here? :)

kevinhartman (approved):

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
jcjimenez: Good point!

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Constants.scala
jcjimenez: lol, sure

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/Constants.scala
kevinhartman: Shouldn't the user need to click an `apply` button before we issue the restart command?

kevinhartman:
Also if you haven't already, make sure to delete the contents of the old checkpoint directory prior to restarting streaming.

kevinhartman:
@jcjimenez I'd also like to request that we delay merging this feature until we've stabilized a bit since making the project more autonomous could add hiccups to the deploy, test, re-deploy loop.

jcjimenez:
OK, I've reduced the scope of this PR to just waiting for `fortis.streams` records to be present.

PR #105, opened 2017-08-23
Remove parallelization of aggregators for UI visibility

kevinhartman:
Removing this unveils previously hidden jobs for our batches, giving us more visibility for performance debugging.


c-w (approved):
Would it make sense for this be configurable via an environment variable so we can easily switch it on/off if we need testing in the future?

kevinhartman:
@c-w I'm somewhat concerned that using `par` here might be unexpected given that the Spark UI does not appear to monitor these threads. After we've optimized the aggregators, additional research and a followup PR can address this properly.

PR #107, opened 2017-08-25
Ensure that a single Cassandra connector is shared

c-w:
As per conversation in Slack. Still needs to be perf-tested to see if there's an improvement. (In any case it shouldn't hurt.)

### DIFF
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
@@ -1,11 +1,13 @@
 package com.microsoft.partnercatalyst.fortis.spark.sinks.cassandra
 
 import java.util.UUID
+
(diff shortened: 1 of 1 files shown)


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/ProjectFortis.scala
```diff
@@ -92,6 +93,7 @@ object ProjectFortis extends App {
 
   // Main starts here
   val ssc = StreamingContext.getOrCreate(fortisSettings.progressDir, createStreamingContext)
+  implicit val connector: CassandraConnector = CassandraConnector(ssc.sparkContext)
```
kevinhartman: Nothing will bind to this here. How about putting this inside `CassandraEventsSink.apply` and making it `lazy` as well (to ensure it doesn't get re-serialized for every task)?

c-w: Done in b889974.

kevinhartman (approved):

PR #109, opened 2017-08-25
Move Cassandra sink helper methods in scope of connection implicit

kevinhartman:
- Moves helper methods in scope of the Cassandra connection implicit.
- Moves registration of UDFs outside of the foreach RDD loop.
- Adds caching for schema-fied fortis events.


jcjimenez (approved):
LGTM

c-w (approved):

PR #116, opened 2017-08-28
Ensuring single topic tuple is always present.

jcjimenez:


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventSchema.scala
```diff
@@ -86,7 +86,7 @@ object CassandraPopularTopics {
 object CassandraConjunctiveTopics {
   def apply(item: Event): Seq[ConjunctiveTopicAggregate] = {
     val keywords = item.computedfeatures.keywords
-    val keywordPairs = if (keywords.size == 1) Seq((keywords.head, "")) else keywords.combinations(2).flatMap(combination => Seq(
+    val keywordPairs = if (keywords.isEmpty) Seq() else keywords.map(k=>(k, "")) ++ keywords.combinations(2).flatMap(combination => Seq(
```
jcjimenez, line 89: if `keywords = ("foo", "bar")` then the tuples will be `( ("foo", ""), ("bar", ""), ("foo", "bar"), ("bar", "foo") )`

erikschlegel (approved):
LGTM

PR #117, opened 2017-08-28
Add metrics for raw time to connect to Cassandra

c-w:
This should be removed once we get to production.


jcjimenez (approved):
LGTM

c-w:
Tested this with a local job and it's a trivial number, so not worth adding to master.

PR #118, opened 2017-08-28
Use same load balancing policy as fortis-services

c-w:
For some reason we're seeing ~80ms write times from fortis-services to Cassandra but multi-second write times from fortis-spark to Cassandra.

One of the differences between the two setups is that fortis-services uses the NodeJS connector which has a different default load-balancing policy than the Spark connector used in fortis-spark which assumes that Spark workers and Cassandra nodes are co-located on the same hosts (which is not the case in our deployment). This change makes the setup consistent so that both projects use the same load-balancing policy.

### DIFF
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraConfig.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraConfig.scala
@@ -15,5 +15,6 @@ object CassandraConfig {
       .setIfMissing("spark.cassandra.auth.username", CassandraUsername)
       .setIfMissing("spark.cassandra.auth.password", CassandraPassword)
       .setIfMissing("spark.cassandra.connection.keep_alive_ms", envOrElse("CASSANDRA_KEEP_ALIVE_MS", (batchDuration.milliseconds * 2).toString))
+      .setIfMissing("spark.cassandra.connection.factory", "com.microsoft.partnercatalyst.fortis.spark.sinks.cassandra.FortisConnectionFactory")
   }
 }
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/FortisConnectionFactory.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/FortisConnectionFactory.scala
@@ -1,0 +1,14 @@
+package com.microsoft.partnercatalyst.fortis.spark.sinks.cassandra
+
+import com.datastax.driver.core.Cluster
+import com.datastax.driver.core.policies.{DCAwareRoundRobinPolicy, TokenAwarePolicy}
+import com.datastax.spark.connector.cql.{CassandraConnectionFactory, CassandraConnectorConf, DefaultConnectionFactory}
+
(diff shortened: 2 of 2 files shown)


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/ConnectionFactory.scala
```diff
@@ -0,0 +1,116 @@
...
+
+import com.datastax.driver.core._
+import com.datastax.driver.core.policies.{DCAwareRoundRobinPolicy, ExponentialReconnectionPolicy, TokenAwarePolicy}
+import com.datastax.spark.connector.cql.CassandraConnectorConf.CassandraSSLConf
+import com.datastax.spark.connector.cql.{CassandraConnectionFactory, CassandraConnectorConf, MultipleRetryPolicy}
+import org.apache.commons.io.IOUtils
+
+/**
+  * Copy-pasted version of DefaultConnectionFactory to over-write the load balancing policy
+  * There is only one change called out by START/END comments below
```
jcjimenez: What if we do something like this instead of copy-pasting so much code?

```
class FortisConnectionFactory extends CassandraConnectionFactory {
  override def createCluster(conf: CassandraConnectorConf): Cluster = {
    val lbp = new TokenAwarePolicy(new DCAwareRoundRobinPolicy.Builder().build())
    DefaultConnectionFactory.clusterBuilder(conf).withLoadBalancingPolicy(lbp).build()
  }
}
```

jcjimenez (approved):
LGTM with question regarding copy-pasted code.

c-w:
Our tests pass but running this in Spark actually crashes:

```
C:\Repos\project-fortis-spark>spark-submit --class CounterDemo --master local[3] target/scala-2.11/project-fortis-spark-assembly-0.0.6.jar
Using Spark's default log4j profile: org/apache/spark/log4j-defaults.properties
Exception in thread "main" java.lang.IllegalArgumentException: Singleton object not available: ConnectionFactory$
        at com.datastax.spark.connector.util.ReflectionUtil$.findGlobalObject(ReflectionUtil.scala:55)
        at com.datastax.spark.connector.cql.CassandraConnectionFactory$$anonfun$fromSparkConf$1.apply(CassandraConnectionFactory.scala:155)
        at com.datastax.spark.connector.cql.CassandraConnectionFactory$$anonfun$fromSparkConf$1.apply(CassandraConnectionFactory.scala:155)
        at scala.Option.map(Option.scala:146)
        at com.datastax.spark.connector.cql.CassandraConnectionFactory$.fromSparkConf(CassandraConnectionFactory.scala:155)
```

This is because for a Scala object (not class), `getClass.getName` appends a dollar sign at the end which leads the class lookup by name to fail.

Fixed this in 2f382aa.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/ConnectionFactory.scala
c-w: Nice one. Fixed in e86952b.

c-w:
This is a ~50% performance improvement over the previous implementation but still orders of magnitude slower than it should be.

PR #120, opened 2017-09-01
Replaced popular places aggregator and updated conjunction topics agg…

jcjimenez:
Local tests:

```
Online aggregator for popularplaces completed in 2335456779 nanoseconds
Online aggregator for popularplaces completed in 1503633332 nanoseconds
Online aggregator for popularplaces completed in 1441249057 nanoseconds
Online aggregator for popularplaces completed in 1129791081 nanoseconds
Online aggregator for popularplaces completed in 1091082498 nanoseconds
Online aggregator for popularplaces completed in 1169926178 nanoseconds
Online aggregator for popularplaces completed in 985164068 nanoseconds
Online aggregator for popularplaces completed in 1088189173 nanoseconds
Online aggregator for popularplaces completed in 857796537 nanoseconds
Online aggregator for popularplaces completed in 1093226156 nanoseconds
Online aggregator for popularplaces completed in 925358995 nanoseconds
Online aggregator for popularplaces completed in 918898127 nanoseconds
```

```
Offline aggregator for PopularPlacesOfflineAggregator completed in 79606246 nanoseconds
Offline aggregator for PopularPlacesOfflineAggregator completed in 73232024 nanoseconds
Offline aggregator for PopularPlacesOfflineAggregator completed in 66754025 nanoseconds
Offline aggregator for PopularPlacesOfflineAggregator completed in 53030312 nanoseconds
Offline aggregator for PopularPlacesOfflineAggregator completed in 63759325 nanoseconds
Offline aggregator for PopularPlacesOfflineAggregator completed in 62546884 nanoseconds
Offline aggregator for PopularPlacesOfflineAggregator completed in 48932904 nanoseconds
Offline aggregator for PopularPlacesOfflineAggregator completed in 62651849 nanoseconds
Offline aggregator for PopularPlacesOfflineAggregator completed in 127617870 nanoseconds
Offline aggregator for PopularPlacesOfflineAggregator completed in 86135597 nanoseconds
Offline aggregator for PopularPlacesOfflineAggregator completed in 64097898 nanoseconds
Offline aggregator for PopularPlacesOfflineAggregator completed in 72830842 nanoseconds
```

PR #121, opened 2017-09-02
Integrating new cassandra tables to support range filters

erikschlegel:


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/aggregators/HeatmapOfflineAggregator.scala
```diff
@@ -0,0 +1,99 @@
...
+
+class HeatmapOfflineAggregator(session: SparkSession) extends OfflineAggregator[HeatmapTile] {
+  private val ComputedTilesTable = "computedtiles"
+
+  override def aggregate(events: RDD[Event]): RDD[HeatmapTile] = {
+    val tiles = events.flatMap(event=>{
+      Seq(
+        event,
+        event.copy(externalsourceid = "all"),
+        event.copy(pipelinekey = "all", externalsourceid = "all")
```
c-w, line 18: Does this also need to include `event.copy(pipelinekey = "all")` so that we can query by an arbitrary pipelinekey without also having to restrict the externalsourceid?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/aggregators/HeatmapOfflineAggregator.scala
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  private def IncrementalUpdateQuery: String = {
+    val GroupedBaseColumnNames = Seq("periodtype", "period", "conjunctiontopic1", "conjunctiontopic2", "conjunctiontopic3", "perioddate", "tileid", "tilez", "pipelinekey", "externalsourceid").mkString(",a.")
+
+    s"SELECT a.$GroupedBaseColumnNames, " +
+    s"       a.mentioncount + IF(IsNull(b.mentioncount), 0, b.mentioncount) as mentioncount, " +
+    s"       a.avgsentimentnumerator + IF(IsNull(b.avgsentimentnumerator), 0, b.avgsentimentnumerator) as avgsentimentnumerator " +
+    s"FROM   aggregatedtiles a " +
+    s"LEFT OUTER JOIN ${ComputedTilesTable} b " +
```
c-w, line 76: Would it be possible to turn the bigint and avgsentimentnumerator in the computedtiles table into counter columns so that we don't have to do this read here? Read-before-write was causing us slowness in the PR that JC just submitted (https://github.com/CatalystCode/project-fortis-spark/pull/120) so we may want to avoid it here too.

Granted, tables with counter columns can't be used in materialized views, but we can also just replace those with normal tables and manage the data replication ourselves by just adding a few more writes from Spark to Cassandra.

Just something to think about.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/aggregators/HeatmapOfflineAggregator.scala
```diff
@@ -0,0 +1,99 @@
...
+      case 0 => return
+      case _ => {
+        implicit val rowWriter = SqlRowWriter.Factory
+        tiles.saveToCassandra(keyspace, "heatmap")
+
+        val tileBuckets = aggregateTileBuckets(tiles, keyspace).cache()
+        tileBuckets.write
+          .format("org.apache.spark.sql.cassandra")
+          .mode(SaveMode.Append)
+          .options(Map("keyspace" -> keyspace, "table" -> ComputedTilesTable)).save
```
c-w, line 95: We've seen this call be up to 3x slower than using saveToCassandra on the RDD. I wonder if we can use the Cassandra connector's native join capability (via [leftJoinWithCassandra](https://github.com/datastax/spark-cassandra-connector/blob/72c4bc6703b71854746d35a8f8cc2108f48f6730/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/RDDFunctions.scala#L175-L198) or [joinWithCassandra](https://github.com/datastax/spark-cassandra-connector/blob/72c4bc6703b71854746d35a8f8cc2108f48f6730/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector/streaming/DStreamFunctions.scala#L112-L140)) on our RDD/DStream to get rid of this save call and stick entirely in the saveToCassandra world (this would also help with consistency with the rest of the sink code).

c-w (approved):

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/aggregators/HeatmapOfflineAggregator.scala
erikschlegel: 100% agree and I had the intention of converting this over to a RDD join. Considering you're OOO, I was going to ask @jcjimenez to run with converting the left outer join from dataframes to RDD. A left outer join is unavoidable in this case as we're unable to use counters as we have too many materialized views that are driven off this table.

c-w: Ok. Just FYI: I'm back in action this week and next :)

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/aggregators/HeatmapOfflineAggregator.scala
erikschlegel: Unfortunately that is not possible. Materialized views can contain any counter columns in Cassandra. We have 3 MVs that are dependent on the `mentioncount ` field.

c-w: Would it be possible to replace the materialized views with straight tables?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/aggregators/HeatmapOfflineAggregator.scala
erikschlegel: We're doing this so we can support GQL queries where either `externalsourceid`  or `pipelinekey` are set to `all`

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/aggregators/HeatmapOfflineAggregator.scala
erikschlegel: not possible as we need the `computedtiles` dataset sorted by `mentioncount` to support our top 5 sources / locations / terms GQL queries.

erikschlegel: Cassandra doesnt allow you to add a counter column to your PK, hence you're unable to sort data on that column

jcjimenez (approved):
LGTM

PR #124, opened 2017-09-05
Remove bad exception for unset shared locations

c-w:
This looks like the results of a bad merge; what reason would we have to throw an exception regarding pageIds when checking the shared locations in a Facebook post? Seems more sensible to simply treat posts without a tagged location as not having shared locations.


jcjimenez (approved):
LGTM

PR #133, opened 2017-09-11
Only keep most granular location for a given name

c-w:
For example, if we have a location name that is both a city and a state (like "New York") we only want to keep the city.

Implementation note: keeping the LocationExtractorFactory's lookup field as a `Map[String, Set[Location]]` although only a single location will be in the set after this change to avoid changing the public API of the factory (i.e., minimize risk surface).

### DIFF
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/CassandraTest.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/CassandraTest.scala
@@ -66,7 +66,7 @@ object CassandraTest {
         title = "twitter post" ),
       analysis = Analysis(
         sentiments = List(.5),
-        locations = List(Location(wofId = "wof-85670485", confidence = Option(1.0), latitude = 12.21, longitude = 43.1), Location(wofId = "wof-85680959", confidence = Option(1.0), latitude = 14.21, longitude = 43.1)),
+        locations = List(Location(wofId = "wof-85670485", layer = "country", confidence = Option(1.0), latitude = 12.21, longitude = 43.1), Location(wofId = "wof-85680959", layer = "country", confidence = Option(1.0), latitude = 14.21, longitude = 43.1)),
         keywords = List(Tag(name = "isis", confidence = Option(1.0)), Tag(name ="car", confidence = Option(1.0))),
        //todo genders = List(Tag(name = "male", confidence = Option(1.0)), Tag(name ="female", confidence = Option(1.0))),
         entities = List(Tag(name = "putin", confidence = Option(1.0))),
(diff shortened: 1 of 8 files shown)


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/CassandraTest.scala
```diff
@@ -107,7 +107,7 @@ object CassandraTest {
           title = "twitter post" ),
         analysis = Analysis(
           sentiments = List(.6),
-          locations = List(Location(wofId = "wof-85670485", confidence = Option(1.0), latitude = 12.21, longitude = 43.1)),
+          locations = List(Location(wofId = "wof-85670485", layer = "coutry", confidence = Option(1.0), latitude = 12.21, longitude = 43.1)),
```
kevinhartman: Typo in country

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/dto/FortisEvent.scala
```diff
@@ -45,6 +46,59 @@ case class Location(
...
+    Location.layerToInt(layer).compare(Location.layerToInt(that.layer))
+  }
+}
+
+object Location {
+  /**
+   * @see https://github.com/whosonfirst/whosonfirst-placetypes
+   */
+  def layerToInt(layer: String): Int = {
+    if ("continent".equalsIgnoreCase(layer)) {
```
kevinhartman: Can you lowercase `layer` first and then do a match expression here?

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/LocationsExtractorFactory.scala
```diff
@@ -15,14 +15,23 @@ class LocationsExtractorFactory(
...
+      val newLocation = toLocation(feature)
+      val oldLocation = map.get(locationName).orNull
+      if (oldLocation == null) {
+        logDebug(s"Got new location for name $locationName: ${newLocation.wofId}")
+        map(locationName) = newLocation
+      } else if (newLocation < oldLocation) {
+        logDebug(s"Discarding location ${oldLocation.wofId} for name $locationName as we now have more granular location ${newLocation.wofId}")
+        map(locationName) = newLocation
+      } else {
+        logDebug(s"Discarding location ${newLocation.wofId} for name $locationName since we already have more granular location ${oldLocation.wofId}")
```
kevinhartman: "Ignoring" may be clearer for ignored incoming locations, especially at a quick glance while debugging (differentiating from a more specific location being found in the logs).

kevinhartman (approved):

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/LocationsExtractorFactory.scala
c-w: 7703962

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/dto/FortisEvent.scala
c-w: a73d5f7

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/CassandraTest.scala
c-w: 94fb918

PR #134, opened 2017-09-11
Dedupe locations/keywords/etc before aggregation

c-w:
In the dashboard we want to show the number of events that match a particular location or keyword, not the number of times that the term was matched in the event.


jcjimenez (approved):
LGTM

PR #135, opened 2017-09-11
Extend blacklist to locations and entities

c-w:
Example situation of when this is useful: currently there is lots of talk about the hurricane in Florida, US. There's also a city called Florida in Colombia, however, that Fortis site wouldn't want to pick up all the events related to Florida, US.


jcjimenez (approved):
LGTM

PR #136, opened 2017-09-11
Sanity check tests + missing equals/hashCode implementations.

jcjimenez:
Found some potential problems around `equals`/`hashcode` that I'm not sure we're hitting, but could pose a problem around redundant counts.

### DIFF
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/Tile.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/Tile.scala
@@ -4,10 +4,7 @@ import com.microsoft.partnercatalyst.fortis.spark.sinks.cassandra.dto.Place
 
 import scala.collection.mutable
 
-class TileId(var tileId: String) {
-  var row: Int = -1
-  var zoom: Int = -1
-  var column: Int = -1
+case class TileId(var tileId: String) {
 
   private val parts = tileId.split('_')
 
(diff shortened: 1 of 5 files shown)


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/Tile.scala
```diff
@@ -4,10 +4,7 @@ import com.microsoft.partnercatalyst.fortis.spark.sinks.cassandra.dto.Place
 
 import scala.collection.mutable
 
-class TileId(var tileId: String) {
-  var row: Int = -1
-  var zoom: Int = -1
-  var column: Int = -1
+case class TileId(var tileId: String) {
```
jcjimenez, line 7: Converting to a case class takes care of `hashCode` and `equals` which was returning `false` for instances with identical `tileId` values.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/transforms/locations/Tile.scala
```diff
@@ -30,6 +30,23 @@ class Tile(var tileId: TileId, var latitudeNorth: Double, var latitudeSouth: Dou
...
+  override def toString = s"Tile($tileId, $latitudeNorth, $latitudeSouth, $longitudeWest, $longitudeEast)"
+
+  def canEqual(other: Any): Boolean = other.isInstanceOf[Tile]
+
+  override def equals(other: Any): Boolean = other match {
+    case that: Tile =>
+      (that canEqual this) &&
+        tileId == that.tileId
+    case _ => false
+  }
```
jcjimenez, line 43: Making sure this returns true for equal values of `tileId` regardless of lat/long values (and precision).

erikschlegel:
LGTM

erikschlegel (approved):
LGTM

PR #143, opened 2017-09-12
Optimize aggregators: partitioning and caching

kevinhartman:
A higher number of partitions was required to use all cores of all executors. Default parallelism is num cores available across all of Spark, but many examples multiply this by 2 (more is better).

Moved event explosion prior to aggregators to avoid recomputing (it can be cached this way).

Current metrics:
screen shot 2017-09-12 at 2 47 03 pm

### DIFF
--- a/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
+++ b/src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
@@ -66,7 +66,16 @@ object CassandraEventsSink extends Loggable {
             writeEventBatchToEventTagTables(eventBatchDF, sparkSession)
           }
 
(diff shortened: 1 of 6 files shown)


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sinks/cassandra/CassandraEventsSink.scala
```diff
@@ -66,7 +66,16 @@ object CassandraEventsSink extends Loggable {
...
           }
 
-          val fortisEventsRDDRepartitioned = fortisEventsRDD.repartition(5)
+          val eventsExploded = fortisEventsRDD.flatMap(event=>{
+            Seq(
+              event,
+              event.copy(externalsourceid = "all"),
+              event.copy(pipelinekey = "all", externalsourceid = "all")
+            )
+          })
```
jcjimenez, line 75: Yes, please!

jcjimenez (approved):
LGTM

PR #145, opened 2017-09-15
Fix regression from capitalization of pipelinekey

kevinhartman:
A requirement from the UI side required `pipelinekey` to become capitalized.


jcjimenez (approved):
LGTM

PR #148, opened 2017-09-18
Performance and memory tuning

kevinhartman:
* Improves computed tiles aggregation performance: now 4 seconds, was 50.
* Fixes a memory leak caused by an RDD not being unpersisted.
  - We're now explicitly unpersisting cached RDDs. Note this may not be required everwhere. A future changeset will remove explicit calls where they're not needed.
  - Note that all `unpersist()` calls are marked blocking to guarantee memory is freed in time with streaming interval. We may be able to relax this in the future for a slight speedup.
* Changes default write behavior to Cassandra to be single write. This has predictable stability, though we should custom tune write behavior for each aggregator in the future.
* Fixes an issue where SB messages sent to the configuration queue without a properties bag would crash the driver.


Smarker (approved):
LGTM

PR #149, opened 2017-09-18
Fix bug preventing entity and sentiment extraction.

kevinhartman:


Smarker (approved):
LGTM

erikschlegel (changes requested):
Check line 69 of Delta.scala. Looks like we're not passing the correct sitesetting.

erikschlegel (approved):

PR #152, opened 2017-09-18
Guarding keyword count to no more than 10.

jcjimenez:


erikschlegel (approved):

kevinhartman (approved):

kevinhartman:
Better to *not* make it configurable through environment variables. If anything, it should be on site settings, right? I would probably leave that out for now.

jcjimenez:
@kevinhartman I think you are right, `sitesettings` would be where to get this from, but for now, let's go with this so we don't have to make any DDL changes.

PR #154, opened 2017-09-19
Fix eventbatch join.

kevinhartman:
Selecting nothing doesn't work. Selecting just the partition key instead for now.


Smarker (approved):
LGTM

PR #160, opened 2017-09-21
Getting RSS entry url from links if uri is not a proper URL.

jcjimenez:


src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/RSSAnalyzer.scala
```diff
@@ -13,21 +13,40 @@ class RSSAnalyzer extends Analyzer[RSSEntry] with Serializable with AnalysisDefa
...
       externalsourceid = item.source.uri,
       pipelinekey = "RSS",
-      sourceurl = item.uri,
+      sourceurl = url.getOrElse(""),
       sharedLocations = List(),
       original = item
     )
   }
 
+  private[analyzer] def getSourceUrlFromItem(item: RSSEntry): Option[String] = {
```
kevinhartman, line 32: Should make these private not package private to avoid people from taking unexpected dependencies.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/RSSAnalyzer.scala
```diff
@@ -13,21 +13,40 @@ class RSSAnalyzer extends Analyzer[RSSEntry] with Serializable with AnalysisDefa
...
+        else getSourceUrl(item.links.head.href)
+      }
+    }
+  }
+
+  private[analyzer] def getSourceUrl(url: String): Option[String] = {
+    try {
+      Option(new URL(url).toURI.toASCIIString)
+    } catch {
+      case _: Exception => None
```
kevinhartman, line 46: What will this throw? Should at least catch nonfatal instead. If you use Try().toOption, catching nonfatal only will be handled automatically :)

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/RSSAnalyzer.scala
```diff
@@ -13,21 +13,40 @@ class RSSAnalyzer extends Analyzer[RSSEntry] with Serializable with AnalysisDefa
...
+      eventid = s"RSS.${url.getOrElse(item.uri)}",
       sourceeventid = item.uri,
       eventtime = item.publishedDate,
       body = body,
       imageurl = None,
       title = item.title,
       externalsourceid = item.source.uri,
       pipelinekey = "RSS",
-      sourceurl = item.uri,
+      sourceurl = url.getOrElse(""),
```
kevinhartman, line 26: Hopefully pipeline doesn't validate this one is non empty? Empty string won't pass it if it's checking this field too.

kevinhartman (approved):

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/RSSAnalyzer.scala
jcjimenez: I don't think it does, but the UI does use this URL. In any case, the absence of this value is pathological/unlikely.

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/RSSAnalyzer.scala
jcjimenez: Well, I was hoping to write some unit tests for these, but it was late =P

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/analyzer/RSSAnalyzer.scala
jcjimenez: Oh, I didn't know about that - will do (separate PR?) To answer your question, this will most often throw something like a java.net.MalformedURLException.

PR #164, opened 2017-09-27
Fix keyword limit enforcement and write batching

kevinhartman:
Previously, title and body keywords were being limited individually to max keywords, allowing for > max keywords when terms do not appear in both title and body.

With this change, a total of max keywords will be selected. Keywords found in the title are always prioritized over those found in the body. The secondary ordering is the number of occurrences.

Max keywords and max locations are now threaded through from environment variables.

Also re-enables batching for writes. We may need to tune these values further.


erikschlegel (approved):
Nice work. Like the word / place limits tunable in the site settings.

PR #166, opened 2017-09-29
Add partitioning controls for streams.

kevinhartman:
**This is needed to take advantage of multiple workers during event analysis for receiver-based DStreams (includes Twitter and RSS).**

Most of our streams are implemented as receivers, so data is not partitioned automatically. This change allows the desired number of partitions for a given stream to be set in its params bundle in Cassandra. If it's omitted, the default behavior is to repartition by Spark's default parallelism (what we want for all cases in existing deployments as of today => no DB changes needed).

For direct DStreams, set 'partitions' to 0 in the stream's params bundle. This will disable repartitioning, defaulting to natural partitioning of the direct stream.


Smarker (approved):
LGTM

PR #167, opened 2017-09-29
Add missing column to computed tiles join columns.

kevinhartman:
Change `joinColumns` value from `PartitionKey` to `PrimaryKey` in computed tiles join.


jcjimenez (approved):
LGTM

PR #170, opened 2017-09-29
Tune Cassandra connection pooling.

kevinhartman:
We may need to tune these further through trial and error, but this should be a good start.

* Increases max # requests per connection to Cassandra binary protocol v3 maximum of 32768.
* Doubles max connections used per node (from 1 to 2).
* Increases heartbeat interval beyond readTimeoutMillis as suggested by Datastax documentation.
* Increases max queue size from 256 to 320 to reduce BusyPoolException during high load.

Reference: http://docs.datastax.com/en/developer/java-driver/3.1/manual/pooling/


jcjimenez (approved):
LGTM - testing locally...

PR #187, opened 2017-10-05
Add shared locations to analysis and remove locations specified in lo…

NathanielRose:
…cation blacklist from events.


jcjimenez (approved):
LGTM

PR #196, opened 2017-10-12
Making default extractLocations look in details.sharedLocations before relying on details.body.

jcjimenez:


c-w (approved):
I wonder if this is the correct approach or if we should use the inferred locations plus the shared locations. Either could work, really depends on what's best for the customer use-cases.

Smarker (approved):
LGTM

PR #197, opened 2017-10-20
Re-enable compilation and test debugger support from IDEA.

kevinhartman:
In order to fix this:
* scalatest was updated to 3.0.4
* scalactic was added as a dependency (required by either? Mockito or scalatest 3.0.4).
* scalatest is now transitively excluded from libraries which incorrectly include it (this allows us to avoid importing the old version as a runtime dependency).


jcjimenez (approved):
LGTM

PR #198, opened 2017-10-25
Handling for > 60 byte watchlist terms in TwitterStreamFactory

kevinhartman:
The Twitter "track" API accepts 400 phrases, where each phrase can have a maximum length of 60 bytes. The resulting stream will contain posts that have been matched by at least one of these phrases. In order for a post to be matched, all terms in the phrase must appear in the post, in any order. The Twitter API splits phrases into terms based on whitespace.

For example, the phrase "fox brown quick" will match a Twitter post with content "The quick brown fox".

**Implementation details:**
* In order to handle phrases > 60 bytes (which aren't supported by track API directly), this implementation splits such phrases into N phrases, minimizing N such that no resulting phrase is > 60 bytes, and minimizing the difference between the number of terms in each phrase.
* If a phrase contains a single `term > 60 bytes`, the corresponding watchlist item is skipped.
* If a phrase splits into N phrases such that `N > 400`, the corresponding watchlist item is skipped.
* If a phrase splits into N phrases such that the `N + |previous track terms| > 400`, and additional Twitter stream must be opened to satisfy the corresponding watchlist item (and any subsequent terms).

The result of this approach is that the resulting stream will contain *at least* all posts that could match the watchlist item (there may be false positives, but these will be eliminated in downstream processing).

*Note: it's possible that a split phrase may include a single common word, such as a stop word. Hopefully, the Twitter track API ignores stop words, but this may not be so. For now, this can be handled on a case-by-case basis.*

**Also includes:**
* Removal of dead userID from params code, which we used to use to specify trusted sources (they now come from the `trustedsources` table in Cassandra).


c-w (commented):
I'm not smart enough to be able to easily follow this code. The explanation in the PR description states the problem well. I wonder if it would be worth to just have a few naive functions to verify each of the pre-conditions for the twitter stream and then splitting it into terms later? That may make it more obvious what's going on.

cicorias:
a question I have in general is where has this been deployed so it can be at least validated it works and no regressions.

jcjimenez (approved):
LGTM

kevinhartman:
While this PR gets the job done, there are some corner cases which could cause trouble around "stop" words. I'd recommend closing this in favor of limiting keywords/phrases to a max of 60 bytes (at the DB level if possible as well as in the admin UI).

c-w:
This is now implemented in project-fortis in https://github.com/CatalystCode/project-fortis/commit/182e2dfb52a53b5b635ef86d3fa1de86e31dd4eb.

image

PR #200, opened 2017-11-06
Add more Application Insights logging

c-w:
We now log dependencies for:
- pipeline.settings (ServiceBus)
- pipeline.sink (Cassandra)
- transforms.entities (OpeNER)
- transforms.featureservice (FeatureService)
- transforms.image (CognitiveService)
- transforms.language (CognitiveService + library)
- transforms.models (AzureBlob)
- transforms.sentiment (CognitiveService + library)

We now log events for:
- pipeline.filters
- batch.receive (same name as previously for backwards compatibility)


jcjimenez (approved):
LGTM

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/StreamsChangeListener.scala
```diff
@@ -116,7 +116,7 @@ object StreamsChangeListener {
           Path(settings.progressDir).deleteRecursively()
         }
       } catch {
-        case e => e.printStackTrace(System.err)
+        case e: Throwable => e.printStackTrace(System.err)
```
kevinhartman, line 119: Best practice is to catch `NonFatal` exceptions since exceptions are used for some control flow in Scala, and catching `Throwable` can actually break the semantics of the language.

```scala
import scala.util.control.NonFatal
...
try { f(x) }
catch { case NonFatal(t) => }
```

src/main/scala/com/microsoft/partnercatalyst/fortis/spark/sources/streamfactories/StreamFactoryBase.scala
```diff
@@ -1,24 +1,18 @@
...
+import com.microsoft.partnercatalyst.fortis.spark.logging.Loggable
 import com.microsoft.partnercatalyst.fortis.spark.sources.streamprovider.{ConnectorConfig, InvalidConnectorConfigException, StreamFactory}
 import org.apache.spark.streaming.StreamingContext
 import org.apache.spark.streaming.dstream.DStream
 
 import scala.reflect.ClassTag
 import scala.util.Try
 
-abstract class StreamFactoryBase[A: ClassTag] extends StreamFactory[A]{
+abstract class StreamFactoryBase[A: ClassTag] extends StreamFactory[A] with Loggable {
```
kevinhartman, line 11: The decorator pattern for logging does not work well for Spark. The problem is that through this approach, the log/telemetry methods become instances methods on the implementing class, meaning that if these methods are used anywhere in functions whose closures are serialized by Spark the entire implementing class will need to be serializable. Today, checkpointing is not enabled, but this change will preclude us from enabling it without making our `StreamFactory`s serializable (which IIRC, is not possible or desirable). 

I understand the desire to have a consistent paradigm for logging and telemetry, but I would recommend rewriting this PR to converge on the old telemetry approach which was chosen originally to address those serialization concerns in a simple way (static / not bound to caller instance).

kevinhartman (changes requested):
Please see my comment in `StreamFactoryBase`. These changes will prevent checkpointing from being supported in the future.

Thanks!
Kevin

c-w:
Closing since this was implemented in https://github.com/CatalystCode/project-fortis/commit/ef557f0628e64ca9af79eeb08b8741768a85588a and https://github.com/CatalystCode/project-fortis/commit/de3a785c8cad806bbf2264a92a924db712bc8f15

